PR #1, opened 2021-04-19
Development since review

thijsmie:
There has been development since the IP review. Also, this pull request adds the documentation not included in the package for IP review.


eboasson (approved):
I've only skimmed the code and, based on that, I really love what this adds to the initial commit. So L(very)GTM!

@k0ekk0ek do what you think is best 🙂

k0ekk0ek:
I'll have a look before the end of the day :+1:

k0ekk0ek:
Since this is all WiP and I'm not the worlds foremost Python expert, I didn't concern myself with the Python binding too much yet. Also, since this is merely to get it up-to-date with the latest/greatest, it's probably best to give it a thorough look/run after the changes have been merged. I'm certain some things will pop-up, but that's why it's still WiP.

@thijsmie, I know you're working on getting Azure Pipelines integration to work [here](https://github.com/thijsmie/cyclonedds-python/tree/pipelines), so I'm not going to require that'd be in this PR. Would be good to have it soon-ish though. Concerning that work, I do like to see Windows and macOS builds.

As for the comments regarding this particular PR:
* There's a mixture of prefixes used in the C code. Using a `ddspy_` prefix for every symbol that's exported is required before we do a release.
* We seem to be using Makefiles to generate documentation. I'm strongly opposed. We have been using CMake for both `cyclonedds` and `cyclonedds-cxx` to not require `make`. CMake is a terrible choice for *just* generating documentation, but I'm sure with a language that runs on all tier-1 platforms we can come up with something "better".
* The `LICENSE` files mention only the `EPL`, I'm pretty sure we want this binding to be available under the `EDL` too, so that's something we should update.
* I've spotted some `malloc`/`free` combo's. It's probably better to use `ddsrt_malloc`/`ddsrt_free` there. They're currently only slightly different on the general purpose operating systems and I very much doubt we'll ever run the Python binding on FreeRTOS, but at some point we might use alternative allocators(?) Perhaps we'll support things like allocation statistics etc. Behavior won't change short-term, but probably better to switch to `ddsrt_` variants in a future PR.
* The IDL generator backend exports some headers and generates a PackageConfig.cmake. That's something you'd do for general purpose libraries, it's best to remove those. The library/plugin will only ever be used by the IDL compiler, perhaps the binding itself too, but never by users directly.

I think it'd be good to update the `LICENSE` files in this PR, the other remarks I don't think require fixing right away.

thijsmie:
Hi @k0ekk0ek! Thanks for the review, I have addressed the LICENCE concern (the EDL was in the main LICENCE but not in the per-package copies), removed the PackageConfig and removed the docs makefile. The makefile is auto-generated by sphinx, but the intended entrypoint to build the documentation is to just run `sphinx-build`. 

As for the `malloc/free` combos: the python interface memory management is on the schedule: it needs to be replaced as part of the post-first-release optimisation push, in which we will then also switch to `ddsrt_malloc/free`.

In terms of exported symbols naming: they are not truly exported since the generated library is internal to the python package, but no harm in making it consistent indeed.

k0ekk0ek:
I'm going to assume that the code in the last commit is correct. It's difficult without going over everything and getting a feel for the binding. Thanks @thijsmie.

PR #4, opened 2021-05-17
Update PyCDR packaging strategy

thijsmie:
PyCDR does not need any complicated setup so this PR updates it to use the newer `setup.cfg` method. `setup.py` is still present as empty shell to allow older (v18 and below) pip distributions to still use the installer.

### DIFF
--- a/src/pycdr/setup.cfg
+++ b/src/pycdr/setup.cfg
@@ -1,0 +1,36 @@
+[metadata]
+name = pycdr
+version = 0.1.5
+author = Thijs Miedema
+author_email = thijs.miedema@adlinktech.com
+description = Python CDR serialization
+url = https://github.com/eclipse-cyclone/cyclonedds-python
+project_urls =
+    Bug Tracker = https://github.com/eclipse-cyclone/cyclonedds-python/issues
+classifiers =
+    Development Status :: 4 - Beta
+    License :: OSI Approved :: Eclipse Public License 2.0 (EPL-2.0)
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
+    Operating System :: OS Independent
+
+[options]
+zip_safe = True
+include_package_data = False
+packages = find:
+python_requires = >=3.6
+install_requires =
+    dataclasses>=0.8; python_version < "3.7"
+    typing-inspect>=0.6; python_version < "3.7"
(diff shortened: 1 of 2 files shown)


src/pycdr/setup.cfg
```diff
@@ -0,0 +1,36 @@
...
+version = 0.1.5
+author = Thijs Miedema
+author_email = thijs.miedema@adlinktech.com
+description = Python CDR serialization
+url = https://github.com/eclipse-cyclone/cyclonedds-python
+project_urls =
+    Bug Tracker = https://github.com/eclipse-cyclone/cyclonedds-python/issues
+classifiers =
+    Development Status :: 4 - Beta
+    License :: OSI Approved :: Eclipse Public License 2.0 (EPL-2.0)
```
eboasson, line 12: Shouldn't EDLv1 be mentioned as well?

eboasson (commented):
LGTM, but I don't have any detailed knowledge about Python packaging, so this is just a common sense thing.

src/pycdr/setup.cfg
thijsmie: It really should, but it is not in the officially supported classifier list. I am looking into how to get it on there.

k0ekk0ek: I wonder if you cannot just use BSD-3-Clause instead? It's the same license and that's what we mention in the headers(?)

thijsmie:
> LGTM, but I don't have any detailed knowledge about Python packaging, so this is just a common sense thing.

The general consensus is to move away from running code as part of the installation process, which is why we get binary wheels for more complex packages and this setup.cfg stuff for simple python-only packages.

src/pycdr/setup.cfg
k0ekk0ek: `SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause` that's what's in the top of our headers...

thijsmie: The full list is [here](https://pypi.org/classifiers/), BSD-3 is also not on there. I am wondering, since these classifiers are just strings, if I can just add new ones. I will run a test on the test-pypi website.

thijsmie: I found instructions in the Pypi FAQ and filed an issue [here](https://github.com/pypa/trove-classifiers/issues/70) for it to be added.

eboasson:
> No activity from Pypi maintainers on the issue so far, not even "you can submit a pull request with this change".  

Bummer ... so

> Shall I just add the "BSD" classifier with a comment that BSD-Clause-3 is intended?

Let's do that.

Issue #5, opened 2021-05-26
on_data_available functor used instead of on_sample_rejected


pbleyer:
https://github.com/eclipse-cyclonedds/cyclonedds-python/blob/334132cdcad1330b766aa408e6290583d93f51bc/src/cyclonedds/cyclonedds/core.py#L747

Should be `self._set_functors["on_sample_rejected"] = self.on_sample_rejected`

thijsmie:
Thank you @pbleyer for spotting this mistake, and also thank you for being the first issue reporter to CycloneDDS-Python!

thijsmie:
@eboasson Text parsing is always hard, "Fix issue #5" was not good enough, that word 'issue' ruins the linking between PR and Issue. Anyway, this is fixed and should be closed.

PR #6, opened 2021-05-26
Fix issue #5

thijsmie:
This functor map is never used for lookups, but purely to avoid Python GC'ing the function if a pointer is passed to the C layer but the original owner of the callable is freed. Thus this mistake can lead, in very rare cases, to a segfault because the C layer will call a freed Python function.


eboasson (approved):
That looks to me like a self-evidently correct fix :)

PR #8, opened 2021-05-31
Keyless semantics

thijsmie:
This PR changes the semantics of `@cdr` to be as follows:
`@cdr` stands for "Keyed topic where all fields are key"
`@cdr(keyless=True)` stands for "Keyless topic, no fields are key"
`@cdr(keylist=[])` stands for "Keyed topic, no fields are key"

Since CI is not merged yet, here is a full run of tests:
https://dev.azure.com/tmiedema/cyclonedds-python/_build/results?buildId=165&view=results

### DIFF
--- a/src/cyclonedds/clayer/src/pysertype.c
+++ b/src/cyclonedds/clayer/src/pysertype.c
@@ -21,6 +21,7 @@
 #include "dds/dds.h"
 
 #include "dds/ddsrt/endian.h"
+#include "dds/ddsrt/heap.h"
 #include "dds/ddsrt/mh3.h"
 #include "dds/ddsrt/md5.h"
 #include "dds/ddsi/q_radmin.h"
@@ -110,6 +111,7 @@ typedef struct ddspy_sertype {
     ddsi_sertype_t my_c_type;
     PyObject* my_py_type;
     cdr_key_vm* key_vm;
+    bool keyless;
     bool key_maxsize_bigger_16;
 } ddspy_sertype_t;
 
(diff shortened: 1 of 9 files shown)


eboasson:
Thanks @thijsmie for such a quick response. I had a quick look at the changes and my impression was that they do what it says on the tin, but that doesn't mean I'm convinced that it says the right thing on the tin:

> This PR changes the semantics of `@cdr` to be as follows:
> `@cdr` stands for "Keyed topic where all fields are key"

That's fine with me, it can be quite convenient and I do think it is one of the two intuitive ways of interpreting the absence of a key specification.

> `@cdr(keyless=True)` stands for "Keyless topic, no fields are key"

Ok ... but ...

> `@cdr(keylist=[])` stands for "Keyed topic, no fields are key"

The definition of a keyless topic is that it has no key fields ... Perhaps I should put it differently: the distinction between `WITH_KEY` and `NO_KEY` exists only in the DDSI-RTPS spec, the actual DDS specification simply talks about topics and their key fields and what happens when the set of key fields for a topic is empty (in that case there can only be a single instance).

The distinction between these two modes isn't necessary in my view. I think keeping only the second form is the best option.

thijsmie:
To be honest, the third option didn't really make sense to me either. I will adapt it to remove the keyless option so the semantics are:
`@cdr` stands for "Keyed topic, all fields are key"
`@cdr(keylist=[])` stands for "Keyless topic, no fields are key"

eboasson (commented):
To the extent that I can review this without really diving into the python implementation, it looks fine. The latest commit you pushed makes me a bit uncertain, so can you please confirm nothing more is coming? If not, I'll merge it, otherwise I'll wait :)

thijsmie:
The commit is finished as-is, it was supposed to be one commit but I made a mistake with git missing a file, which I added in a new commit instead of amending and force pushing

PR #9, opened 2021-06-01
Remove variable group Default reference.

thijsmie:
While my setup uses variable groups the CycloneDDS-Python Azure DevOps instance does not. This PR simply removes the variable group from the pipelines.


k0ekk0ek:
Thanks @thijsmie!

PR #10, opened 2021-06-02
Qos Reliability BestEffort Policy update

thijsmie:
This PR removes the Reliability BestEffort max_blocking_time parameter. This parameter is only used for Reliability.Reliable and serves only as a nuisance for BestEffort.


eboasson (approved):
Apologies for having let this sit for so long!

PR #11, opened 2021-06-03
Fix Codecov pipeline

thijsmie:
When adding Windows Debug builds to the main pipeline I did not re-check the codecov pipeline, and it turns out that is now broken. This PR addresses the issue.

### DIFF
--- a/.azure/publish-coverage.yml
+++ b/.azure/publish-coverage.yml
@@ -30,7 +30,7 @@ variables:
 - name: python-version
   value: 3.9
 - name: build_type
-  value: RelWithDebugInfo
+  value: RelWithDebInfo
 
 steps:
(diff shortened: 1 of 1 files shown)


.azure/publish-coverage.yml
```diff
@@ -29,11 +29,13 @@ pool:
 variables:
 - name: python-version
   value: 3.9
+- name: build_type
+  value: RelWithDebugInfo
```
k0ekk0ek, line 33: If this is supposed to be the CMake build type, shouldn't it be RelWithDebInfo?

thijsmie: You know, it should be and I changed it, although the pipeline and tests didn't seem to care...

PR #14, opened 2021-06-23
Fix Python Qos repr()

thijsmie:
This removes the sorting of policies in the `__repr__` of the python Qos object, since the policies are already sorted upon creation of the Qos object. This re-sorting didn't specify how to sort the objects and could fail, causing a crash. A test is also added to ensure the repr(qos) always works. Parametrized Qos tests are also combined for readability.


eboasson (approved):
Thanks @thijsmie !

PR #15, opened 2021-06-23
Documentation pipeline

thijsmie:
This adds automatic documentation generation in the same style as [Eclipse Cyclone DDS does](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/.azure/guides.yml). We will need to add the "comsumer" side of this [in the website repository](https://github.com/eclipse-cyclonedds/eclipse-cyclonedds.github.io/blob/master/azure-pipelines.yml) at some point too.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -49,7 +49,7 @@ pygments_style = 'friendly'
 
 # -- Allow documentation building without loading libraries -------------------
 
-os.environ['CDDS_NO_IMPORT_LIBS'] = "1"
+os.environ['CYCLONEDDS_PYTHON_NO_IMPORT_LIBS'] = "1"
 typing.TYPE_CHECKING = True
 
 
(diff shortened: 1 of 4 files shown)


eboasson (approved):
LGTM, but I do not actually feel qualified to judge whether this is as it should be. @k0ekk0ek if you could look it over, too, I'd be grateful.

docs/source/conf.py
```diff
@@ -51,4 +44,13 @@
...
 
 html_theme = 'sphinx_rtd_theme'
-pygments_style = 'friendly'
\ No newline at end of file
+pygments_style = 'friendly'
+
+
+# -- Allow documentation building without loading libraries -------------------
+
+os.environ['CDDS_NO_IMPORT_LIBS'] = "1"
```
k0ekk0ek, line 52: Can we change this to `CYCLONEDDS_NO_IMPORT_LIBS` to avoid confusion? I know it's nitpicking, it was already in there, but I prefer using the actual project name consistently...

k0ekk0ek (changes requested):
Just the one little nitpick, other than that I'm absolutely fine with this

thijsmie:
That should do it then :)

eboasson:
> That should do it then :)

Yep 🙂

PR #17, opened 2021-06-24
Coverage inside Azure Pipelines

thijsmie:
This pull request switches from codecov to coverage inside Azure Pipelines. An example run can be seen [here](https://dev.azure.com/tmiedema/cyclonedds-python/_build/results?buildId=313&view=codecoverage-tab).


k0ekk0ek (approved):
I like it :+1: Coverage per run is great. Could we do something like having the build fail if coverage drops?

thijsmie:
I think [this](https://marketplace.visualstudio.com/items?itemName=mspremier.BuildQualityChecks) would work for that. Of course it would also be possible to fail the CI based on a fixed threshold from inside the pipeline.

eboasson (approved):

eboasson:
I wonder what's up with the ECA check ...

k0ekk0ek:
Eclipse had an outage yesterday, I hit revalidate and it's green now. Merging.

PR #21, opened 2021-06-29
Change to user installs with pip in CI

thijsmie:
As demonstrated by [recent CI runs](https://dev.azure.com/eclipse-cyclonedds/cyclonedds-python/_build/results?buildId=593&view=logs&j=8e372324-b069-5441-ff7c-d552df30aa05&t=fd7072a0-3888-59bc-79c3-7952cda91f57) the default `pip install` path on windows is now a protected location. This commit changes all `pip install` invocations to use a `--user` flag. It also adds pip caching back in.
Because the `--user` install location is not on PATH on macOS by default this then breaks the mac build, so modules need to be invoked through Python instead of directly.

### DIFF
--- a/.azure/publish-documentation.yml
+++ b/.azure/publish-documentation.yml
@@ -43,8 +43,8 @@ steps:
   - bash: |
       set -e -x
       cd docs
-      pip install --upgrade pip
-      pip install -r requirements.txt --user --upgrade
+      pip install --user --upgrade pip
+      pip install --user --upgrade -r requirements.txt
     name: install_sphinx
     displayName: Install Python dependencies
   - bash: |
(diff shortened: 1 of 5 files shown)


thijsmie:
Well that was a journey. Running a tool in a subprocess and sending ctrl-c is trivial on linux and macos but turned out to be a massive headache. In the end I solved it by not starting the tool through a shell but running it inside python in a thread.

eboasson:
> massive headache on windows

One could be forgiven for thinking "they" enjoy inflicting [cruel and unusual punishment](https://en.wikipedia.org/wiki/Cruel_and_unusual_punishment) on their users ... on the flip side, all the greater the thankfulness when someone braves that (censored)!

PR #24, opened 2021-07-16
Some compiler fixes and switch to scikit-build

k0ekk0ek:
Was trying something and wanted to compile against build install. Thought the scikit-build was a nice addition as it allows for better integration with CMake, which is used for Cyclone DDS. i.e. when users use the Windows installer it'll find everything without having to set `CYCLONEDDS_HOME`. I'll have to see how to find the proper path for runtime too, so it's a draft. We probably want to have a look at the fact `ddspy` is now a module as opposed to an extension.

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -38,6 +38,8 @@ steps:
       pip install --user ./src/cyclonedds
     name: install_cyclonedds_py
     displayName: Run installers
+    env:
+      cflags: -Werror
   - bash: |
       python -m flake8 ./src/pycdr/pycdr --count --select=E9,F63,F7,F82 --show-source --statistics
       python -m flake8 ./src/cyclonedds/cyclonedds --count --select=E9,F63,F7,F82 --show-source --statistics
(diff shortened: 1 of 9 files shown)


k0ekk0ek:
@thijsmie, I think this is ready for review. It works with the basic build & test pipeline, not sure if the publish-release pipeline works though. Anything special I need to set there?

k0ekk0ek:
Fixed a bunch of compiler warnings too. Should be good to go(?)

thijsmie:
I spotted the update and ran the release pipeline. This needs a quick fixup: CMAKE_PREFIX_PATH should be set or the wheel builds [error out](https://dev.azure.com/tmiedema/cyclonedds-python/_build/results?buildId=418&view=logs&j=1d1ee225-4ba4-5094-8c7d-890c0a92be80&t=5d58c467-e016-505a-d5f3-be5fe4bcc8ee). Also the `cflags=-werror` would be nice to have in the release pipeline too.

Fixing the C warnings was overdue, thank you for tackling them, changes all look sensible. The very few python changes are also fine.

PR #25, opened 2021-07-21
Fixes, Scoped QoS, Windows cmake registry and release pipeline fixes

thijsmie:
Collection of some updates, grouped per commit:
* First commit is a collection of some fixes, mostly discovered during tutorial writing. It also adds scoped QoS for use in pubsub and support for Property and BinaryProperty Qos.
* Second commit adds a dds library loader for windows by peeking the windows registry for cmake installs.
* Third commit fixes the release pipeline for use with scikit-build.

### DIFF
--- a/src/cyclonedds/cyclonedds/core.py
+++ b/src/cyclonedds/cyclonedds/core.py
@@ -11,11 +11,13 @@
 """
 
 import uuid
+import asyncio
+import concurrent
 import ctypes as ct
 from weakref import WeakValueDictionary
 from typing import Any, Callable, Dict, Optional, List, TYPE_CHECKING
 
-from .internal import c_call, c_callable, dds_c_t, DDS
+from .internal import c_call, c_callable, dds_infinity, dds_c_t, DDS
 from .qos import Qos, Policy, _CQos
 
 
@@ -1509,6 +1511,20 @@ class WaitSet(Entity):
         ret = self._waitset_set_trigger(self._ref, value)
         if ret < 0:
             raise DDSException(ret, f"Occurred when setting trigger in {repr(self)}")
+
+    async def wait_async(self, timeout: Optional[int] = None) -> int:
+        """Asynchronously wait for a WaitSet to trigger. Use in event-loop based applications.
+        
+        Parameters
+        ----------
+        timeout: int, Optional = None
+            Maximum number of nanoseconds to wait before returning. By default this is infinity.
+        """
+        timeout = timeout or dds_infinity
+
+        loop = asyncio.get_running_loop()
+        with concurrent.futures.ThreadPoolExecutor() as pool:
+            return await loop.run_in_executor(pool, self.wait, timeout)
 
     @c_call("dds_create_waitset")
     def _create_waitset(self, domain_participant: dds_c_t.entity) -> dds_c_t.entity:
--- a/src/cyclonedds/cyclonedds/pub.py
+++ b/src/cyclonedds/cyclonedds/pub.py
@@ -10,11 +10,13 @@
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 """
 
-from typing import Optional, TYPE_CHECKING
+from typing import Optional, Union, TYPE_CHECKING
 
 from .internal import c_call, dds_c_t
-from .core import Entity, DDSException
-from .qos import _CQos
+from .core import Entity, DDSException, Listener
+from .domain import DomainParticipant
+from .topic import Topic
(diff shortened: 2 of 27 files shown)


src/cyclonedds/cyclonedds/core.py
```diff
@@ -1510,6 +1512,13 @@ def set_trigger(self, value: bool) -> None:
         if ret < 0:
             raise DDSException(ret, f"Occurred when setting trigger in {repr(self)}")
 
+    async def wait_async(self, timeout: Optional[int] = None) -> int:
+        timeout = timeout or 0xFFFFFFFFFFFFFFFF
```
k0ekk0ek, line 1516: The `0xFFF...`, where does that come from? Does that mean infinity/no timeout? There's probably some constant that's better used in this case.

src/cyclonedds/cyclonedds/pub.py
```diff
@@ -29,20 +31,36 @@
...
     def __init__(
             self,
-            domain_participant: 'cyclonedds.domain.DomainParticipant',
-            qos: Optional['cyclonedds.core.Qos'] = None,
-            listener: Optional['cyclonedds.core.Listener'] = None):
+            domain_participant: DomainParticipant,
+            qos: Optional[Qos] = None,
+            listener: Optional[Listener] = None):
+        if not isinstance(domain_participant, DomainParticipant):
+            raise TypeError(f"{domain_participant} is not a cyclonedds.domain.DomainParticipant or cyclonedds.sub.Subscriber.")
```
k0ekk0ek, line 38: Telling the user that he can also pass in a `cyclonedds.sub.Subscriber` seems weird?

src/pycdr/pycdr/types.py
```diff
@@ -62,95 +62,164 @@ def __repr__(self):
...
+        else:
+            return 406820761152609 ^ hash(self.subtype)
+
+    __str__ = __repr__
 
 
-class BoundStringHolder:
-    def __init__(self, max_length):
-        self.max_length = max_length
+class bound_str(metaclass=GetItemSupportMeta):
```
k0ekk0ek, line 142: Can we perhaps just call it `bounded_string`? This hurts my eyes :slightly_smiling_face:

k0ekk0ek (changes requested):
I'm really not proficient enough in Python to do a proper review of this, so I'll have to give it a try. Did manage to find some minor things and really like the release pipeline. Also, do we have tests that try to create a publisher from another entity etc, that would've caught the subscriber thingy.

src/cyclonedds/cyclonedds/core.py
thijsmie: Ah yes I had planned to implement and use `cyclonedds.util.duration(infinite=True)` but this would cause a circular import. I will add `dds_infinity` to `internal.py` and implement that both in `duration` and here in the waitset.

src/cyclonedds/cyclonedds/pub.py
thijsmie: Copy-paste failure, will fix.

src/pycdr/pycdr/types.py
thijsmie: I can, but the Python datatype is called `str` and not `string`. How about `bounded_str`?

k0ekk0ek: Sounds good to me :+1:

src/pycdr/pycdr/types.py
```diff
@@ -62,95 +62,164 @@ def __repr__(self):
...
-        self.max_length = max_length
+class bounded_str(metaclass=GetItemSupportMeta):
+    @classmethod
+    def __class_getitem__(cls, tup):
+        if type(tup) != tuple:
+            tup = (tup,)
+
+        if len(tup) != 1 or type(tup[0]) != int:
+            raise TypeError("A bounded str takes one argument, a maximum length.")
+        if tup[0] <= 0 or tup[0] > 18446744073709551615:
```
k0ekk0ek, line 150: Where does the large constant come from? That's probably like a `SIZE_MAX` or something? Wouldn't it be way easier if we'd get that from the C layer? I know they're defines not symbols, but adding a `const` to the `clayer` that we can use here wouldn't be too much work right? Same for things like DDS_INFINITY etc... Would that be an option?

thijsmie: I think I could actually remove this check, if you try to set a string limit of what in quick estimation are several exabytes then it will go wrong in other places.
I will move DDS_INFINITY into the `clayer`, I believe I can easily add it with `PyModule_AddIntMacro`.

thijsmie:
Superseeded by #33.

PR #26, opened 2021-07-21
CDR Key Vm Alignment test&fix

thijsmie:
Simple testcase that verifies alignment works correctly. This will segfault while the [cdr keyvm fix](https://github.com/YixianLv/cyclonedds-python/commit/c4e791afba245d9f2e515e0f0bd2058b4b7fe923) is not in.


k0ekk0ek:
Does it make sense to have the fix in this PR too?

thijsmie:
Sure thing @k0ekk0ek, fix is now included.

k0ekk0ek:
My bad. Thanks @thijsmie!

PR #33, opened 2021-08-18
One package, Idl&Serializer verification

thijsmie:
This pretty big pull request combines the pycdr and cyclonedds packages into one cyclonedds package. Secondly, a lot of work has gone into verification of the output of the IDL python backend, merging it into the package and adding a discovery mechanism for `idlc`. Next some fuzzy testing of the serializer revealed errors in both the python and c implementation.

The motivation for combining the two packages is as follows: the types for python lived in pycdr, but the idl backend is part of cyclonedds. This creates a tightly coupled versioning nightmare (the longer you think about it the more hairy it becomes). Now everything lives under `cyclonedds.idl` and you could still use it in a standalone way if really needed.

This pull request superseedes https://github.com/eclipse-cyclonedds/cyclonedds-python/pull/25, that made it easier to avoid conflicts.


thijsmie:
Segfault on windows fixed, running procdump on CI was painful. Also tracked down the source of the previous windows test flakyness and fixed it (race condition within test wrt temporary directories).

eboasson (approved):
I had taken a look a while ago, when the Windows CI was still failing. Except for those failing tests, it looked good to me then already. Of course one must allow for the possibility that my enthusiasm was enabled by a lack of familiarity with some of the details, though that is by no means certain 🙂 In any case, now that it passes, it has my approval!

I'll wait with merging it for another day or so in case @k0ekk0ek would also want to have a chance to look at it.

k0ekk0ek:
I'm fine with the changes. Thijs and I discussed packaging idlc with the Python package extensively, so I'm definitely in favor of that. The other changes all seem fine :+1:

k0ekk0ek (approved):

PR #34, opened 2021-08-19
Relax test on status changes

eboasson:
Needed because of status mask behaviour fix
https://github.com/eclipse-cyclonedds/cyclonedds/commit/a3ac2fd83c2a087cd01d2118b7ff8070fe4b8984

Signed-off-by: Erik Boasson <eb@ilities.com>


thijsmie:
Indeed a necessary change, it was already included in #33 😀

PR #46, opened 2022-01-07
XTypes support

thijsmie:
It's here, complete XTypes support for Python! This includes features that the C library doesn't yet support (strongly connected components, arbitrary key content) so those cannot be fully tested yet.
The testsuite will segfault on dynamic_subscribe tests until [cyclonedds#1099](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1093) is merged. There might also be one or two 'Mutated python did not communicate with C' fuzzer failures, they are false positives, as the 'poor-mans-assignabilty-checking' in the python testsuite needs some more attention.


eboasson (approved):
With the updated Cyclone, CI is as expected. For anyone who thinks it is implausible that I could have reviewed this PR in detail already, you're correct. I have been keeping an eye on it, including on the quality of the testing, and at this point and with my limited knowledge of python, there's not much point in further reviewing.

Besides which, as far as I can tell, it works. Thanks @thijsmie for this lovely New Year's present!

PR #47, opened 2022-01-10
Fix for failing mutation test in fuzzing

thijsmie:
Makes sure that there can't be any empty appendable structs and is a bit more careful with mutating things inside final unions, which should fix any remaining fuzzing failures.


eboasson (approved):
LGTM!

PR #49, opened 2022-01-18
Topic generic typing

thijsmie:
This pull request doesn't introduce any new functionality, it just sorts out the type hints for topics, datareaders and datawriters so that the .read* en .write* methods will give correct type information to static type checkers, making them much more pleasant to use in an IDE. Consider the following:

```python
from cyclonedds import domain, topic, sub
from cyclonedds.idl import IdlStruct

class A(IdlStruct):
    f1: int
    f2: float

dp = domain.DomainParticipant()
tp = topic.Topic(dp, "bla", A)
dr = sub.DataReader(dp, tp)
a = dr.read()[0]
print(a.|
```
I have introduced a mock cursor. Previously IDE's would not be aware that typeof(a) == A so you would not get the autocomplete options "f1" or "f2". This pull request fixes this.

### DIFF
--- a/cyclonedds/__init__.py
+++ b/cyclonedds/__init__.py
@@ -9,3 +9,19 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 """
+
+from . import internal, util, qos, core, domain, topic, pub, sub, builtin, dynamic, idl
+
+__all__ = [
+    "internal",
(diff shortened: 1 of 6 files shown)


eboasson (approved):
That sounds lovely, and even to this Python-noob it makes sense that these changes have this effect 😁

PR #50, opened 2022-01-19
Some minor fixes on idlpy

thijsmie:
On windows the lib was not installed to the correct folder, since a 'dll' qualifies as a runtime artifact not a library like a '.so'. Not that the windows lib works now, the segfault due to free'ing memory from another library (thus the need for idl_free) is still present.

Also fixed a minor issue where the interaction between 'py-root-prefix' and types in the top level would result in a faulty import.


eboasson (approved):
LGTM, CI test failures are unrelated as far as I can tell.

PR #51, opened 2022-01-24
Fix several issues with untyped serdatas

thijsmie:
As it turns out there were several mistakes in the SDK_KEY implementation, by which keys without CDR header could be added into messages or not properly aligned. It is fixed here by never allowing a keyed bit of data to exist without header in front of it.


thijsmie:
Took a few tries to get the details right, but it should be good now :)

PR #53, opened 2022-01-26
Enable bounded sequences in fuzzer types

eboasson:
Now that the C code supports bounded sequences, it seems only sensible to also include them here. Not sure whether 8 is the right value to use, I suspect that means it is now generating more sequences. I did have a look at some generated types, and I don't think they have excessively many sequences.

Signed-off-by: Erik Boasson <eb@ilities.com>

### DIFF
--- a/tests/support_modules/fuzz_tools/rand_idl/containers.py
+++ b/tests/support_modules/fuzz_tools/rand_idl/containers.py
@@ -38,7 +38,7 @@ class RTypeDiscriminator(Enum):
             8,
             8,
             8,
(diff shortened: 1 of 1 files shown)


thijsmie:
LGTM! Types are indeed properly generated, confirmed locally. (nice that that worked first try). With the passing testsuite it seems that indeed the bounded sequences are working properly.
As for the probabilities, we can tweak these if needed, but looking at the generated idl I think it looks okay for now.

eboasson:
Thanks @thijsmie 🙂

PR #61, opened 2022-03-01
Add missing colons to if statements.

pbleyer:
Typos I guess.


thijsmie:
Hi there! You are of course right, these are missing. Now I am confused about a few things: I must have overlooked it in a review, which okay, that happens. However, the fact that there is not a single test that caught this (nor a hard error by flake8) is concerning...

eboasson:
Solved by #62

PR #62, opened 2022-03-02
Fix typo and add tests to catch these.

thijsmie:
Replaces pull request #61, thanks @pbleyer for reporting it! This PR also adds a test to catch these fundamental mistakes.


eboasson (approved):
Thanks @thijsmie (and @pbleyer)

PR #69, opened 2022-03-11
Set allowed data representation

thijsmie:
Following PR https://github.com/eclipse-cyclonedds/cyclonedds/pull/1188 here the fixes for python I promised to @dpotman.


eboasson (approved):
CI failures are noise.

PR #76, opened 2022-03-31
Removed several assert() statements in idlpy context.c file 

torsten-pf:
On Windows (MSVC, debug build), several asserts are triggered when idlc is called for Python. However, these do not seem to be necessary or valid. I removed them and the generator works for debug builds as well.
By the way, I had to manually build the generator plugin with debug information, since setup.py only generates release builds. And the debug version of idlc does not work with the release build of the plugin.

### DIFF
--- a/idlpy/src/context.c
+++ b/idlpy/src/context.c
@@ -180,7 +180,6 @@ void idlpy_ctx_free(idlpy_ctx octx)
     assert(octx->basepath);
     assert(octx->idl_file);
     assert(octx->module != NULL);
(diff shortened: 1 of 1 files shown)


thijsmie:
Hi @torsten-pf,

Thank you for your contribution! I've pulled these lines into another PR since I had to fix some other IDL related bugs and ran into the same issues in a debug build. We will also backport the fixes to a 0.9.1 release. Let me know if there is anything else you run into!

PR #82, opened 2022-05-06
Fix path-to-idlc in wheel mistake.

thijsmie:
I tested the `idlc` that is delivered with the python wheels and it turns out it doesn't work quite right. This PR should go in master and be backported to a 0.9.1 release, which I'll submit after code review.

You can test by hotpatching this fix into the python wheel:
```python
import cyclonedds
from pathlib import Path
p = Path(cyclonedds.__file__).parent / 'tools'/'wheel_idlc'/'__init__.py'
p.write_text(p.read_text().replace(
  "libdir = Path(__file__).resolve().parent / '.libs'",
  "libdir = Path(cyclonedds.__file__).resolve().parent / '.libs'"
).replace(
  "os.execvpe(idlc, sys.argv[1:], environ)",
  "os.execvpe(idlc, sys.argv, environ)"
))
```

### DIFF
--- a/cyclonedds/tools/wheel_idlc/__init__.py
+++ b/cyclonedds/tools/wheel_idlc/__init__.py
@@ -21,7 +21,7 @@ import cyclonedds
 from pathlib import Path
 
 
-libdir = Path(__file__).resolve().parent / '.libs'
+libdir = Path(cyclonedds.__file__).resolve().parent / '.libs'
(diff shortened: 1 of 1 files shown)


eboasson (approved):
CI failures surely must be unrelated but let's see what #86 shows.

thijsmie:
And github rerun is still borked but another manual rerun is green: https://dev.azure.com/eclipse-cyclonedds/cyclonedds-python/_build/results?buildId=2714&view=results

PR #86, opened 2022-05-10
Small fixes for CI, sync with C

thijsmie:
This has become a little bit of a hodgepodge of changes, but at least they are still grouped by commit!

 * Commit 015deb8 syncs the TRY_CONSTRUCT flags with C PR https://github.com/eclipse-cyclonedds/cyclonedds/pull/1242
 * Commit 2191b9e fixes a leak in the python sertype where a `ddsi_sertype` does not get unref'ed when a topic creation fails.
 * Commit 672f459 fixes a small mistake in TypeObjects of unions with an Enum as discriminator
 * Commit 43e9b2c fixes a mistake that only occurs on windows where the C api can return NULL from an entity name while claiming it is present. This is dealing with the symptom though, I am not exactly sure why it occurs (it doesn't reproduce locally for me, only on CI...)

This _should_ make the CI fully green again, provided the fuzzer doesn't find anything random 😄

### DIFF
--- a/cyclonedds/idl/_xt_builder.py
+++ b/cyclonedds/idl/_xt_builder.py
@@ -390,6 +390,17 @@ class XTBuilder:
             _ctype = toscan.pop()
             my_node_name = _ctype.__idl_typename__.replace('.', '::')
 
+            if isclass(_ctype) and issubclass(_ctype, IdlUnion):
+                # get_extended_type_hints will not inspect the discriminator, and that can be an enum
+                discriminator_type = _ctype.__idl_discriminator__
+                if isclass(discriminator_type) and issubclass(discriminator_type, IdlEnum):
+                    scan_node_name = discriminator_type.__idl_typename__.replace('.', '::')
+                    if scan_node_name not in graph:
+                        graph[scan_node_name] = set()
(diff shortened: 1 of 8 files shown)


cyclonedds/idl/_xt_builder.py
```diff
@@ -27,6 +25,9 @@
 from ._type_normalize import get_extended_type_hints, get_idl_annotations, get_idl_field_annotations, WrapOpt
 
 
+uint32_max = 2 ** 32 - 1
```
eboasson, line 28: I don't Python all that well, so forgive me for asking: is this going to be fine on a 32-bits platform?

eboasson (approved):

cyclonedds/idl/_xt_builder.py
thijsmie: Python integer types are effectively BigInt implementations with unlimited precision, so you can keep raising powers till you run out of memory 😄.

thijsmie:
@eboasson the CI on master had a failure in the fuzzer due to a case in a union being called `get` which name-clashed with a `IdlUnion` method, an interesting case indeed 😅.

PR #88, opened 2022-05-12
Update to 0.9.1

eboasson:
Collected fixes + a bump in the version number:

* re-opening modules in IDL
* getting the right path to IDLC in the wheel
* accessing of the `IS_OPTIONAL` annotation in the type objects
* encoding of the `try_construct` annotation if left at the default
* memory leak on topic creation failure
* handling of enums as union discrimant


thijsmie (approved):
Looks good to me, and with https://dev.azure.com/eclipse-cyclonedds/cyclonedds-python/_build/results?buildId=2727&view=results the CI is also happy!

PR #91, opened 2022-05-19
New cli tools

thijsmie:
This adds a new suite of command line tools to the python backend:

```sh
# list the entities discovered on the network
$ cyclonedds ls
# list the applications discovered on the network (once participant-properties are in)
$ cyclonedds ps
# discover and reconstruct the IDL of a topic type
$ cyclonedds typeof topic_name
# discover and subscribe to an arbitrary topic
$ cyclonedds subscribe topic_name
```

### DIFF
--- a/cyclonedds/tools/cli/discovery/main.py
+++ b/cyclonedds/tools/cli/discovery/main.py
@@ -1,0 +1,243 @@
+import time
+import re
+from typing import List
+from cyclonedds import core, domain, builtin, dynamic, util
+from datetime import datetime, timedelta
+from collections import defaultdict
+
+from ..utils import LiveData
+from ..idl import IdlType
+from .ls_discoverables import DParticipant, DTopic, DPubSub
+from .ps_discoverables import PApplication, PParticipant, PSystem
+
+
+def ls_discovery(live: LiveData, domain_id: int, runtime: timedelta, topic: str) -> List[DParticipant]:
+    try:
+        topic_re = re.compile(f"^{topic}$")
+    except re.error:
+        topic_re = re.compile(f"^{re.escape(topic)}$")
+
+    dp = domain.DomainParticipant(domain_id)
+
+    rdp = builtin.BuiltinDataReader(dp, builtin.BuiltinTopicDcpsParticipant)
+    rcp = core.ReadCondition(rdp, core.SampleState.NotRead | core.ViewState.Any | core.InstanceState.Alive)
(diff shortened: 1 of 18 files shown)


cyclonedds/tools/cli/discovery/main.py
```diff
@@ -0,0 +1,243 @@
...
+    address_get = core.Policy.Property("__NetworkAddresses", "")
+
+    start = datetime.now()
+    end = start + runtime
+    while datetime.now() < end and not live.terminate:
+        for p in rdp.take(N=20, condition=rcp):
+            if p.key == dp.guid and not show_self:
+                continue
+
+            hostname = p.qos[hostname_get].value if p.qos[hostname_get] is not None else "Unkown"
```
eboasson, line 145: "Unkown" is probably be a typo, and copied to the following three lines

eboasson (approved):
Neat! From what I can tell there's no hard dependency on https://github.com/eclipse-cyclonedds/cyclonedds/pull/1281, so no need to wait for it.

That means the only comments I have are a typo and wondering how the color scheme works out for people using a light background. I should try that out myself ... I know.

Neither is a reason to not approve this already. Thanks @thijsmie !

cyclonedds/tools/cli/discovery/main.py
thijsmie: Good catch, fixed that.

thijsmie:
Indeed there is no hard dependency on the participant-properties, given that we can't rely on them always being there anyways it would not be sensible to require them for the tool to work.

As for the color scheme, I have no idea! I have left basically all settings of the wonderful [click](https://github.com/pallets/click/) by @pallets, [rich](https://github.com/Textualize/rich) by @willmcgugan and [rich-click](https://github.com/ewels/rich-click) by @ewels libraries on default and I have good hope that that will 'just work'! I'll see if I can manage a testrun by fiddling with my terminal settings.

eboasson:
@thijsmie, if `pip install --user --upgrade pip` now suddenly fails on Windows, I'd say it is unrelated to this PR. As the CI was fine on all platforms until I merged https://github.com/eclipse-cyclonedds/cyclonedds/pull/1281 and neither that nor the changes you made to this PR in response are platform dependent, I'd say the likelihood of this PR introducing any problems on Windows is negligible. As far as I am concerned, this can be merged right now.

thijsmie:
Yeah I don't know what goes on with windows either, I will merge now and hopefully the problem with pip will go away, if not I'll see about changing the way the windows python environment is set up.

PR #93, opened 2022-05-27
Fix the reuse of typedefs (and sequence/array types).

### LINKED ISSUE #85: Creating topics from dynamically discovered types fails with a typedef array
I'm getting the following error when I attempt to instantiate a topic using a dds type that was dynamically discovered. I believe that the only change I made recently was to use typedefs in the idl for arrays.

```
topic = Topic(self._domain_participant, topic_name, topic_type)
  File "cyclonedds/topic.py", line 54, in __init__
    data_type.__idl__.populate()
  File "cyclonedds/idl/_main.py", line 106, in populate
    self.v0_machine, self.v2_machine, self.keyless, self.version_support = Builder.build_machines(self.datatype)
  File "cyclonedds/idl/_builder.py", line 311, in build_machines
    v0_machine, v2_machine, keyless = cls._machine_struct(_type)
  File "cyclonedds/idl/_builder.py", line 200, in _machine_struct
    v0_members = {
  File "python_deps_cyclonedds/cyclonedds/idl/_builder.py", line 201, in <dictcomp>
    name: cls._machine_for_type(field_type, False, False)
  File "cyclonedds/idl/_builder.py", line 178, in _machine_for_type
    raise TypeError(f"{_type} is not valid in IDL classes because it cannot be encoded.")
TypeError: typing.Any is not valid in IDL classes because it cannot be encoded.
```

The IDL looks something like the following

```
typedef double foo__double__3[3];

struct Foo {
  foo__double__3 foo_array;
};

```

Any help would be wonderful!

thijsmie:
Fixes #85. Test that reproduces issue `test_dynamic_repeated_typedef_array` was added, fails with current release.

### DIFF
--- a/cyclonedds/idl/_xt_builder.py
+++ b/cyclonedds/idl/_xt_builder.py
@@ -1397,7 +1397,7 @@ class XTBuilder:
 class XTParseState:
     class Deferred:
         def __init__(self):
-            self.callback = None
+            self.callbacks = []
 
     def __init__(self, database):
         self.database: Dict[xt.TypeIdentifier, xt.TypeObject] = database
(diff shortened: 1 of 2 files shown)


eboasson (approved):
LGTM!

Issue #96, opened 2022-06-01
unable to install cyclonedds python package on ARM64 : cyclone_search module not found


zam1kor:
on ARM64 machine, when I install cycloneDDS python package I get the following error:
- I have also followed the instructions as per the [installation from source](https://github.com/eclipse-cyclonedds/cyclonedds-python#installing-from-source), and still I get the same error i.e., cyclone_search module not found

![image](https://user-images.githubusercontent.com/14125075/171444384-e072914e-d3d1-410f-8d94-a3f2d67d4555.png)

zam1kor:
above mentioned issue is solved by installing `python3-dev` package. Details can be found at #98 

FYI  I had followed below steps to install CycloneDDS in ARM64 ubuntu image
```
apt-get update
apt-get install -y cmake

git clone https://github.com/eclipse-cyclonedds/cyclonedds
cd cyclonedds && mkdir build install && cd build
cmake .. -DCMAKE_INSTALL_PREFIX=../install
cmake --build . --target install
cd ..
export CYCLONEDDS_HOME="$(pwd)/install"

apt-get install python3-dev

pip3 install --user "cyclonedds[dev] @ git+https://github.com/eclipse-cyclonedds/cyclonedds-python"
```

PR #97, opened 2022-06-02
Add MANIFEST.in 

thijsmie:
Turns out the source distribution pushed to pypi is broken for all releases so far, since files used in building the package are not included by default. The MANIFEST.in file specifies which additional files need to go into the source distribution. This fixes issue #96 once we make another release. This should also be cherry-picked into the 0.9.x branch.


eboasson (approved):
I'll take your word for it 🙂

thijsmie:
Fuzzer test fail is a false-positive with IDLC failing on a case-sensitivity thing in the test preparation.

PR #99, opened 2022-06-15
Make a new command line tool `cyclonedds performance`

thijsmie:
The new CLI tool `cyclonedds performance` serves as an attempt to *wow* a new user by fancy terminal graphics, courtesy of Rich with some of my own plotting magic. The TUI does not cover all info as provided by `ddsperf` but sometimes a quick visual is all that is needed.

This PR also has screenshots of all the CLI tools as svg files, courtesy of @ewels secret project we'll eventually run in CI. See [the readme on my branch](https://github.com/thijsmie/cyclonedds-python/tree/ddsperf-cli#command-line-tooling) for a visual.


eboasson (approved):
LGTM! 🤩🥳

cyclonedds/tools/cli/ls.py
```diff
@@ -1,27 +1,56 @@
...
+    help="Show the tools own participant and subscriptions.",
+)
+@click.option(
+    "--suppress-progress-bar",
+    type=bool,
+    is_flag=True,
+    help="Suppress the output of the progress bar",
+)
+@click.option(
+    "--force-color-mode",
```
NoxPardalis, line 40: Do we also respect the `NO_COLOR` environment variable for those who don't wish to see any colors in their terminal output? See https://no-color.org/.

thijsmie: Indeed, `rich` supports the `NO_COLOR` variable, which means this is not an issue purely due to the fact of choosing awesome libraries to implement things you have no clue about :innocent:

cyclonedds/tools/cli/sub.py
```diff
@@ -3,54 +3,89 @@
...
+    help="Duration of discovery scan.",
+)
+@click.option(
+    "--suppress-progress-bar",
+    type=bool,
+    is_flag=True,
+    help="Suppress the output of the progress bar",
+)
+@click.option(
+    "--force-color-mode",
```
NoxPardalis, line 37: I was reflecting on this option somewhat. I feel as if it would be more appropriate for this to be a `--color` flag which takes an enumeration of `auto`, `always`, or `never`. The behavior of `--force-color-mode` would  then be achieved by  `--color always`.

NoxPardalis (changes requested):
I have highlighted (what I think is) a misspelled command-line option and I have one suggestion with respect to the `--force-color-mode` flag. Other than that, this all looks great! 👍

NoxPardalis (approved):
The new changes look good to me!

PR #101, opened 2022-06-16
Missed update in processing last review comments.

thijsmie:
In processing the last review comment I missed two success-criterium/criterion renames and made it non-functional. This fixes that.


eboasson (approved):
With this change, `cyclonedds performance` works for me. Yay!

Issue #31, opened 2021-08-02
Listener notification thread hangs while removing DataWriter


jarekdr:
Hi,
while the python try to remove `DataWriter` object within the listener notification thread, the thread hangs. 
Below is a simple program to reproduce it. First `DataWriter` linked with `response_topic` is created, and then `DataReader` linked with `alive_topic` within the same process. So the reader and writer are linked with different topics, and reader has set listener for `on_liveliness_changed`. Next the simple publisher program (running in the separate process) create a `DataWriter` linked with `alive_topic`. When the publisher process stops, `on_liveliness_changed` event in the first process arise, and when there are no more matched writers, `DataWriter` is removed from `list` in the listener method. Since reference to `DataWriter` was only on the `list`, python try to remove that object. But when it calls `__del__` in `core.py`, it get stucked at `def _delete(self, entity: dds_c_t.entity)`

Main program
```
list = []
class CustomListener(Listener):

    def on_liveliness_changed(self, reader, status) -> None:
        if status.alive_count == 0:
            list.pop(0)
            return


dp = DomainParticipant(0)
publisher = Publisher(dp)

response_topic = Topic(dp, 'response_topic', Message, qos=QOS)
list.append(DataWriter(publisher, response_topic))

alive_topic = Topic(dp, 'alive_topic', Message, qos=QOS)
reader = DataReader(dp, alive_topic, listener=CustomListener())

sleep(120)
```

Simple publisher

```
    dp = DomainParticipant(0)
    publisher = Publisher(dp)
    topic = Topic(dp, 'alive_topic', Message, qos=QOS)
    writer = DataWriter(publisher, topic)

    for i in range(1,5):
        print(".")
        writer.write(Message('.'))
        sleep(1)
```

Moreover in the provided example the `on_livenliness_changed` callback was used, but the same situation happens when removing is in the `on_data_available`, regardless publisher already sent or not samples. Reading the issue https://github.com/eclipse-cyclonedds/cyclonedds-python/issues/7,  it was mentioned that call writers is not prohibited in callback until it doesn't concert the same entity. Here I have different topics, so the question is what else could be wrong here?

Thanks in advance

thijsmie:
Hi @jarekdr, thanks for your report.

I've verified your problem with a reproducer both in Python and in C, I will have to do some more investigation as to how the deadlock (or mutual conditional variable wait I think) state comes to be.
[...]

PR #103, opened 2022-06-28
Listener docs addendum + autoformat

thijsmie:
In issue #31 @jakecdr mentioned that the listener limitations and gotcha's should be mentioned in the class docstring, which is a good point. I also took the opportunity to run `core.py` through the `black` autoformatter, which I plan to slowly adopt as standard whenever I edit files as to not make one massive whitespace commit.


eboasson (approved):
👍🏻

PR #111, opened 2022-07-21
[docs] Remove reference to an `idl` decorator which (no longer?) exists.

willstott101:
Small doc change - I can't find an idl decorator mentioned anywhere else... so I assume it was an in-progress thing.


thijsmie (approved):
The CI failure is unrelated, seems to be an Azure glitch. Thanks for the contribution!

PR #112, opened 2022-07-22
__idlc__ must be invoked to print the shared library location

willstott101:
Importing __idlc__ has no side effects - it must be executed as __main__

If anyone has advice on how to fix this legal boilerplate let me know... I think I have signed everything - but then the bot is still complaining without a descriptive error afaict

### DIFF
--- a/docs/source/idl.rst
+++ b/docs/source/idl.rst
@@ -14,7 +14,7 @@ You use the IDL compiler if you already have an idl file to define your types or
 
 .. code-block:: shell
 
-    idlc -l $(python -c "import cyclonedds.__idlc__") your_file.idl
+   idlc -l py your_file.idl
 
 
 .. _datatypes:


thijsmie:
Ah yes this is out of date documentation, but your proposed update is still not the "optimal" way to invoke nowadays since `idlc` is aware of this trick and you can just type `idlc -l py` and let it determine the library location.

As for the legal stuff, I think there has to be a matching email between git commits and eclipse, or the github username must be filled in on your eclipse account? I am not quite sure which one of the two matters or if it is both.

willstott101:
Ah yes, I'll update - I was trying to use `-l python` which fails with an unfound .so error, making me assume it was a virtualenv thing.

I think I have all three - my eclipse account has my github username listed, my email listed, and the commit appears to match. The only idea I has is my "legal name" doesn't match my git name, but that seems ridiculous to me.

willstott101:
Oh it passed, cool. Thanks if anyone had to do anything manual for that.

thijsmie (approved):
Thanks for the contribution!

PR #113, opened 2022-07-22
Explain read/take DataReader's docstrings

willstott101:
I'm not entirely sure my descriptions here are correct (knowledge from https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/AccessingData_Samples_ReadTake.htm via google).

I feel as though there are likely more QoS settings that affect the queue - I'd like to exhaustively list them here if you know where I can find what that list is?

### DIFF
--- a/cyclonedds/sub.py
+++ b/cyclonedds/sub.py
@@ -150,6 +150,10 @@ class DataReader(Entity, Generic[_T]):
         """Read a maximum of N samples, non-blocking. Optionally use a read/query-condition to select which samples
         you are interested in.
 
+        Reading samples does not remove them from the :class:`DataReader's<DataReader>` receive queue. So read
+        methods may return the same sample in multiple calls. Your :class:`History<cyclonedds.qos.Policy.History>`
+        QoS Policy controls how "old" a sample can be whilst remaining in the queue.
+
         Parameters
         ----------
         N: int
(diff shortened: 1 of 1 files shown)


thijsmie:
Hi @willstott101!

I'm really appreciating that you are contributing to the documentation! Maybe you want to join the CycloneDDS Discord and we can talk in the python channel? That allows for some more direct collaboration. 

In any case, the History Policy doesn't really control how "old" a sample can be, since timing has no influence on it. It is purely how many samples lookback you have, which is then in turn subject to the ResourceLimits policy. If you want to control the maximum age of samples the Lifespan is of relevance. I think documenting all behaviour of policies in the python docstring is not the right strategy here: some centralized documentation on how Policies work is probably more suitable. A standard list can be found in the OMG wiki (https://www.omgwiki.org/ddsf/doku.php?id=ddsf:public:guidebook:06_append:02_quality_of_service:start) and in the DDS spec. Maybe you would like to start on a guide for https://cyclonedds.io ? We can then link to it from the docstrings. Or maybe you have a better idea?

willstott101:
Yeah I'll hop in discord - thanks for the links! Perhaps it will indeed be too complex to get across neatly in a docstring - particularly if the description applies to multiple places (publishers & subscribers perhaps). As long as we can link to the guide from the docstring I'm happy.

willstott101:
Closing this in favour of #123 and a future addition of a guide/high-level QoS and Dataspace doc section

PR #114, opened 2022-07-22
Build topic/type discovery into cyclonedds core in quick-start instructions.

willstott101:
Quite a small readme change but ought to save some people confusion.


thijsmie:
I'm not sure about adding this, because on one side I agree, on the other side this will be the default behaviour starting from the next release. I guess it cannot hurt to mention explicitly?

willstott101:
I see, fair enough. I noticed that it's explicitly stated that the pypi version enables this, so thought it'd make sense to add here. If not necessary moving forward then keeping the commands clear probably for the best.

PR #118, opened 2022-07-27
Release pipeline source build fix

thijsmie:
Well, it almost worked! Turns out the pipeline did not have permissions to push to "cyclonedds-nightly" so we have some permissions to add (will send those over @eboasson) but it did get permissions to push to the main package, which it promptly did with the source distribution which I forgot to update to say "cyclonedds-nightly". I've retracted the erronious release and I pushed the builds to PyPi-cyclonedds-nightly now manually so I can generate access tokens, this fix needs to be in before the next release can build (I've disabled the pipeline until then)


eboasson (approved):
LGTM!

PR #123, opened 2022-08-03
Some general doc string improvements

willstott101:
* Added some docs to DataWriter
* Added subset of #113 to DataReader
* Generally tried to improve annotations about timestamps
* Entity links & types fixed to reduce sphinx complaints

Worth asking if you're on board with trying to delegate to type hints (and type-hint syntax) in the docs as much as possible? Seems to me it'll be easier to keep in sync - however for the life of me I can't figure out how to annotate the return type with a description and still have sphinx pull the annotation from the function. So return types are still duplicated in the docstrings sadly.

### DIFF
--- a/cyclonedds/core.py
+++ b/cyclonedds/core.py
@@ -1752,7 +1752,7 @@ class WaitSet(Entity):
         Parameters
         ----------
         timeout: int
-            The maximum number of nanoseconds to block. Use the function :func:`duration<cdds.util.duration>`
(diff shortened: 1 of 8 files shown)


willstott101:
Also fixed the examples int he Introduction document

thijsmie (approved):
Thank you so much @willstott101! This is greatly helping to improve the general quality of the docs.

PR #124, opened 2022-08-05
Add missing _set_on_data_on_readers method to Listener

willstott101:
I should probably write a test for this... but opening MR now whilst I have it done

Noticed when using this:
```python
from cyclonedds.core import Listener

def make_print_listener(name, listener=None):
    def printer(evt_name):
        def fn(*args, **kwargs):
            arg_string = ", ".join(repr(a) for a in args)
            if args and kwargs:
                arg_string += ", "
            arg_string += ", ".join(f"{k}={v!r}" for k, v in kwargs.items())
            print(f"{name!r}.{evt_name}({arg_string})")
        return fn

    l = Listener(
        on_data_available=printer("on_data_available"),
        on_inconsistent_topic=printer("on_inconsistent_topic"),
        on_liveliness_lost=printer("on_liveliness_lost"),
        on_liveliness_changed=printer("on_liveliness_changed"),
        on_offered_deadline_missed=printer("on_offered_deadline_missed"),
        on_offered_incompatible_qos=printer("on_offered_incompatible_qos"),
        on_data_on_readers=printer("on_data_on_readers"),
        on_sample_lost=printer("on_sample_lost"),
        on_sample_rejected=printer("on_sample_rejected"),
        on_requested_deadline_missed=printer("on_requested_deadline_missed"),
        on_requested_incompatible_qos=printer("on_requested_incompatible_qos"),
        on_publication_matched=printer("on_publication_matched"),
        on_subscription_matched=printer("on_subscription_matched"),
    )
    if listener:
        l.merge(listener)
    return l

```

### DIFF
--- a/cyclonedds/core.py
+++ b/cyclonedds/core.py
@@ -1329,6 +1329,12 @@ class Listener(DDS):
     ) -> None:
         pass
 
+    @c_call("dds_lset_data_on_readers")
+    def _set_on_data_on_readers(
+        self, listener: dds_c_t.listener_p, callback: _data_on_readers_fn
+    ) -> None:
+        pass
+
     @c_call("dds_lset_sample_lost")
     def _set_on_sample_lost(
         self, listener: dds_c_t.listener_p, callback: _on_sample_lost_fn
--- a/cyclonedds/core.py
+++ b/cyclonedds/core.py
@@ -936,6 +936,10 @@ class Listener(DDS):
             listener.setters[name](functor)
 
     def merge(self, listener: "Listener") -> None:
+        """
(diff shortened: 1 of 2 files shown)


thijsmie:
Hi @willstott101,

Good catch! Would you be willing to add basically the code snippet you posted as a test? It is indeed good to verify that all the callbacks are correctly registered.

willstott101:
Ok, added a test that exercises both the keyword args and the set_on_* methods.

I would normally avoid such a complicated test but since the root of the bug was tedious code I didn't want to make adding a new event name tedious here too :stuck_out_tongue: (although I guess that's exceedingly rare in a standardised project like this).

thijsmie:
Thank you for your again nice contribution! 😄

PR #127, opened 2022-08-10
Document SampleInfo

willstott101:
I realised I had no idea how to access `source_timestamp` - once digging learned I could do with using `instance_handle` too.

I'm not 100% sure about my docs for the `*_state` attributes - are there more possible values than those in the linked enums? Since those enums are for input really, I might be wrong about it.

### DIFF
--- a/cyclonedds/internal.py
+++ b/cyclonedds/internal.py
@@ -247,6 +247,12 @@ class SampleInfo:
 
     Attributes
     ----------
+    sample_state:
+        Possible values: :class:`SampleState<cyclonedds.core.SampleState>`
+    view_state:
+        Possible values: :class:`ViewState<cyclonedds.core.ViewState>`
+    instance_state:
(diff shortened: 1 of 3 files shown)


thijsmie:
This looks right to me! It is equivalent with the [C SampleInfo](https://cyclonedds.io/docs/cyclonedds/latest/api/sub.html#c.dds_sample_info). This does touch on one of the items on my list of *backwards-compatibility-breaking-changes*: I want to change the InstanceState/ViewState/SampleState to not be magic integer constant (combining with the | operator is debatable but interpreting the constants in the debugger is disgusting, give me readable stuff :innocent:). It is now also too easy to write down illegal masks, InstanceState.Any should be the default etc. Maybe an `enum.IntFlag` or some custom magic is required. Instance handles should also be objects and not integers.

It is a shame we don't have [Intersection types](https://github.com/python/typing/issues/18) so we could add the `sample_info` as a proper type hint via an intersection of the datatype and a protocol. If you have ideas there I'd be happy to discuss.

PR #128, opened 2022-08-10
Now really set the runtime correctly

thijsmie:
That serves me right for making a quick change and thinking it would be correct just because it was only a couple characters! Now it should be okay (and I properly tested it now...).


eboasson (approved):
Looks good (just like the first attempt 😁)

PR #130, opened 2022-08-18
Fixing segfault with reading from BuiltinTopicDcpsTopic

willstott101:
Mainly opening the MR to check if it reproduces on the CI - I consistently get segfault when using this builtin topic - which I wanted to use for type introspection of known topic names.

### DIFF
--- a/clayer/pysertype.c
+++ b/clayer/pysertype.c
@@ -1856,9 +1856,6 @@ ddspy_read_topic(PyObject *self, PyObject *args)
 
     for(uint32_t i = 0; i < ((uint32_t)sts > Nu32 ? Nu32 : (uint32_t)sts); ++i) {
         PyObject *type_id_bytes = NULL;
-        if (rcontainer[i] == NULL) {
-            continue;
-        }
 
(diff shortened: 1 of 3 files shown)


willstott101:
The `dds_builtintopic_endpoint` array `rcontainer` in `ddspy_read_endpoint` in `pysertype.c` appears to have invalid entires? Valgrind is reporting invalid reads when `rcontainer[i]->qos` is accessed... even when checking if  the qos is `NULL`.

I'm a little confused here, and would need to learn a lot to dig further. So I'm going to leave this and avoid the introspection for a while, hopefully will find a chance to come back to this.

willstott101:
Ahah! In the cyclonedds C tests the Topic endpoint is of type `dds_builtintopic_topic_t` not `dds_builtintopic_endpoint_t`: https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/src/core/ddsc/tests/topic_discovery.c#L139

So I think we need ANOTHER read/take pair for the topics in `pysertype.c`

thijsmie (approved):

thijsmie:
Thanks for your work on this @willstott101! :partying_face:

PR #131, opened 2022-08-22
Maybe fix the hang in CI?

willstott101:
From my testing the hang seems to be during cyclonedds Entity deletion. I think it must be related to freeing twice as this MR helps.

However I haven't confirmed this is what's happening, so not really that happy with this.

In other news my wider application test suite now segfaults with this change :) So we still have some way to go. It's up to you if you'd rather back out of the circular reference support at all until I've figured this out...

### DIFF
--- a/cyclonedds/core.py
+++ b/cyclonedds/core.py
@@ -190,8 +190,12 @@ class Entity(DDS):
         if not hasattr(self, "_ref"):
             return
 
-        self._entities.pop(self._ref, None)
-        self._delete(self._ref)
+        try:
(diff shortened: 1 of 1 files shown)


willstott101:
Ok, with my morning less-dead brain this appears to simply be a revert of https://github.com/eclipse-cyclonedds/cyclonedds-python/pull/129/commits/54b90025c2f0c1b8d95a8acd10c9581f00b5aa3e so no point in merging this it'll just be a confusing changelog. Best to revert the other commit.

I have an idea about a solution - will do some further testing.

PR #136, opened 2022-08-25
Release GIL when creating topics

willstott101:
Fixes (at least 1 of) the deadlocks discussed in #134

It doesn't fix the deadlock triggered by https://github.com/eclipse-cyclonedds/cyclonedds-python/commit/54b90025c2f0c1b8d95a8acd10c9581f00b5aa3e sadly, so circular references still can't be supported.

### DIFF
--- a/clayer/pysertype.c
+++ b/clayer/pysertype.c
@@ -975,7 +975,9 @@ ddspy_topic_create(PyObject *self, PyObject *args)
 
     if (sertype == NULL) return NULL;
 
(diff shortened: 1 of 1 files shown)


thijsmie:
Thanks @willstott101 for comfirming this fixes your deadlock. I think we will fully change the listener to not call directly into python but to schedule async notifications via that Python C API I mentioned. That means the listener calls are fully decoupled between Python and C and you can also ignore calls on listeners that are already freed on the python side.

PR #144, opened 2022-09-07
force to run with python3

ThijsSassen:
When trying to create a package with cpack on a red hat based system i get the following error

*** ERROR: ambiguous python shebang in local-ci.py: #!/usr/bin/env python. Change it to python3 (or python2) explicitly.

as the python api needs python 3 a change from python to python3 will fix this error.


thijsmie:
Thanks for the contribution 😃

PR #146, opened 2022-09-12
Emit raw string for generated `library_path`

### LINKED ISSUE #145: `unicodeescape` error in `cyclonedds.__library__` when building on Windows
I was doing some tests on running the `cyclonedds` CLI tools with [pyoxidizer](https://pyoxidizer.readthedocs.io/en/stable/index.html). When attempting to build the executable I get the following error.
```
error[PYOXIDIZER_PYTHON_EXECUTABLE]: adding PythonExecutable to FileManifest

    Caused by:
        0: building Python executable
        1: building executable with Rust project
        2: obtaining embedded python context
        3: converting cyclonedds.__library__ to resource
        4: compiling in-memory bytecode
        5: compiling error: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape (cyclonedds.__library__, line 2)
```

This seems to be because my `dll` for Cyclone is on my path under `\Users` and `\U` begins a unicode escape in Python. For reference, the generated `__library__.py` module on my systems is as follows:
```py
in_wheel = False
library_path = 'C:\Users\noxpardalis\eclipse\cyclonedds\install\bin\ddsc.dll'
```

NoxPardalis:
This prevents issues under Windows where the generated path may contain a `\U` which begins a unicode escape sequence. Emitting a raw string prevents those sequence from being interpreted in the generated `__library__.py` module.

Closes #145

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ if "BUILDING_SDIST" not in os.environ:
 
     with open(this_directory / 'cyclonedds' / '__library__.py', "w", encoding='utf-8') as f:
         f.write("in_wheel = False\n")
-        f.write(f"library_path = '{cyclone.ddsc_library}'")
+        f.write(f"library_path = r'{cyclone.ddsc_library}'")
 
 
     ext_modules = [


thijsmie:
Hi @NoxPardalis, good catch! Indeed backslashes in the path can cause illegal strings.

willstott101:
should this be `f.write(f"library_path = {cyclone.ddsc_library!r}")`? that should delegate all escaping to python

thijsmie:
Oh @willstott101 that is a new one for me and indeed somewhat more elegant. I'll leave this for now though, since it "works" but if I have to touch this again for another reason I'll replace it. (or you can submit a PR and I'll approve it :innocent:)

PR #147, opened 2022-09-13
Change how union equality is determined

thijsmie:
XTypes specifies the following:

_Two types are equivalent according to the Minimal equivalence relation if and only if either
they have equal Fully-Descriptive TypeIdentifiers, or else they have equal Minimal TypeIdentifiers._

I happily checked `is_fully_descriptive(typeid1) and typeid1 == typeid2` where typeid1 are `TypeIdentifier`s. These are `IdlUnion` in python. However, for primitive type kinds the discriminator is set to an *unmapped* discriminator, falling back to the default case, which it doesnt have, so it contains "nothing/void".

Now my "clever" code for union equality assumed that two unions both in the default case with equal value were equal. Clearly that cannot be the case, I simply was not as clever as the XTypes spec in coming up with the idea of using unmapped labels for unions for something else, silly me!

The new behaviour is that two discriminator values always result in non-equality. This can bite someone when they use `None` and `0` for a discriminator interchangeably but that is the cost of following the XTypes spec.

### DIFF
--- a/tests/test_unions.py
+++ b/tests/test_unions.py
@@ -1,0 +1,14 @@
+from cyclonedds.idl import IdlUnion, types
+
+
+class A(IdlUnion, discriminator=types.uint8):
+    f1: types.case[1, types.uint8]
+    f2: types.default[types.uint16]
+
+
+def test_union_not_equal():
+    # Even though these both map to the default case
+    # The very "thoughtful" XTypes API maps these to type kinds without case
+    # Which is very nice...
+    assert A(discriminator=3, value=1) != A(discriminator=2, value=1)
+
--- a/cyclonedds/idl/__init__.py
+++ b/cyclonedds/idl/__init__.py
@@ -149,10 +149,7 @@ class IdlUnion(metaclass=IdlUnionMeta):
         if self.__class__ != other.__class__:
             return False
         if self.discriminator != other.discriminator:
-            if self.discriminator is None and other.discriminator != self.__idl_default_discriminator__:
-                return False
-            if other.discriminator is None and self.discriminator != self.__idl_default_discriminator__:
-                return False
+            return False
         if self.value != other.value:
             return False
         return True


tests/test_unions.py
```diff
@@ -0,0 +1,14 @@
...
+class A(IdlUnion, discriminator=types.uint8):
+    f1: types.case[1, types.uint8]
+    f2: types.default[types.uint16]
+
+
+def test_union_not_equal():
+    # Even though these both map to the default case
+    # The very "thoughtful" XTypes API maps these to type kinds without case
+    # Which is very nice...
+    assert A(discriminator=3, value=1) != A(discriminator=2, value=1)
```
eboasson, line 13: The way I understand the description of the problem, it is that equality of union types is implemented differently from what the XTypes spec says. Since the spec may well be found to be perfect should the universe first be subjected to sufficient cutting-and-pasting, the code may indeed need a bit of work to also be perfect in that altered universe.

However, this test makes me think it is testing whether two different instantiations of a single type (`A`) are indeed considered different, and that's something else. I'm sure I'm misinterpreting something ... XTypes has a habit of making one do that, and my Python skills aren't all that great either.

So I need a little bit of help 🙂

thijsmie: We sort of have in python what we have in javascript, there is equality (`==`) and there is equivalence (`is`). So you can say that `1 is 1` -> `True` and `[1,2,3] == [1,2,3]` -> `True` but `[1,2,3] is [1,2,3]` -> `False` since they are two different lists thus not the same object. The test here is testing for equality and would not succeed without this patch, since the old code considered them to both map to the `default` case with equal value so they were `equal`. I could add different unit tests that test equality and equivalence for some different setups, I did not feel it was needed 😅

eboasson: Thanks! There's of course still the issue of the fuzzers failing, but that seems to be related to https://github.com/eclipse-cyclonedds/cyclonedds/pull/1397 (we really need to figure out how to run C/C++/Python CI in one go ...) and unrelated to this.

eboasson (approved):

PR #150, opened 2022-09-16
Add PyOxidizer support

thijsmie:
This PR allows PyOxidizer to build a self-contained binary for the `cyclonedds` cli tool. Some trickery was added to the `__library__.py` generated file to allow the resulting binary to dynamically locate the CycloneDDS library when installed using common mechanisms like `CYCLONEDDS_HOME`, `LIBRARY_PATH` or install location. Thanks to @NoxPardalis for the idea and minimal prototype which was my starting point.


buildhelp/oxidize_library.py
```diff
@@ -0,0 +1,137 @@
...
+    # Locate relative to current binary
+    dir = good_directory(Path(sys.executable).parent.parent)
+    if dir:
+        return dir
+
+    if "CYCLONEDDS_HOME" in os.environ:
+        dir = good_directory(Path(os.environ["CYCLONEDDS_HOME"]))
+        if dir:
+            return dir
+    if "CycloneDDS_ROOT" in os.environ:
```
eboasson, line 99: This one looks wrong, other than it looks fine to me!

thijsmie: Good catch, this mistake is actually also present in another part of the code, I will fix both instances.

eboasson (approved):
LGTM!

PR #153, opened 2022-10-10
Appendable enums

thijsmie:
While @appendable enums worked properly in Python, the Python IDL backend did not emit the annotation so it would be treated as final.


dpotman (approved):
LGTM!

Issue #105, opened 2022-07-04
idl enum generator


nate800:
The idl generator allows for a python type to be generated from the following definition.
```
enum example {
  None,
  Another
}
```

When generating python code, None is a keyword.

thijsmie:
Hi @nate800,

I do agree with you this is an issue. I am having some trouble with how to fix this without destroying assignability in XTypes, because if you change member names or literal names that check will fail. Some "shadow administration" with fake vs real member names will be necessary. For now I recommend you "just work around it" and wait till either I get around to go through the effort of reworking the name administration or someone else does.

eboasson:
IIRC, somewhere in the IDL spec it says a leading underscore is used to work around keywords. Something like a leading underscore is stripped off from symbols on input, and prepended to any symbol that clashes with a keyword in the target language.

I suppose being masters of the universe, the OMG need not worry about the possibility of a language disallowing a leading underscore and all kinds of other fun things. Or perhaps I have not read the spec carefully enough to have noted that way this transformation is done is language dependent, with the leading underscore simply being the one used for C/C++.

javiersorianoruiz:
Hello, eboasson, I don't know if I have understood you well.

I have tried this IDL code:

```
module vehicles
{
    enum example {
      _None,
      Another
    };

    struct Vehicle
    {
        string name;
        int64 x;
        int64 y;
    };
};

```

When I compile it with IDLC for python I get this code (notice that it is generated without the leading underscore). This code it is not going to work for the None keyword:

```
class example(idl.IdlEnum, typename="vehicles.example", default="None"):
    None = auto()
    Another = auto()
```

When I compile it with IDLC for C I get this code:

```
typedef enum vehicles_example
{
  vehicles_None,
  vehicles_Another
} vehicles_example;

```

If we follow the C philosophy for python, we wouldn't have the problem with the keyword None, we could generate with IDLC for python code like this:

```
class example(idl.IdlEnum, typename="vehicles.example", default="None"):
    vehicles_None = auto()
    vehicles_Another = auto()
```

Could this solution be acceptable?

thijsmie:
Hi @javiersorianoruiz,

While this proposition works in your example, there are some problems with it:

 * What about enums in the global namespace (e.g. not in a module). They still suffer the same problem with keywords.
 * It is not "developer-ergonomic" to force the user to type long strings, especially if there are 15 nested modules.

I see the real solution as follows:

[...]

Issue #161, opened 2022-11-08
idlc python output name conflicts with cyclonedds.idl


JacobJBeck:
Using `idlc -l py` creates a module named 'idl'. This is problematic because within that module, generated scripts have these lines:
```
import cyclonedds.idl as idl
import cyclonedds.idl.annotations as annotate
import cyclonedds.idl.types as types

# root module import for resolving types
import idl
```

This causes `AttributeError: partially initialized module 'idl' has no attribute 'IdlEnum' (most likely due to a circular import)` due to the overlap in names. Renaming the module and replacing the line `import idl` with the new name resolves the issue.

thijsmie:
Yeah that is indeed something I hadn't considered. @javiersorianoruiz how is your keyword filter coming along? We should also add `idl`, `annotate` and `types` as reserved identifiers....

thijsmie:
This should be fixed on the master branch.

PR #163, opened 2022-11-11
Reserved Python keywords support (Issue 105)

javiersorianoruiz:
We (@Bit-Seq, @jordialexalar and me) have implemented a solution for the issue 105.

https://github.com/eclipse-cyclonedds/cyclonedds-python/issues/105

With this change, now, the IDL file can contain reserved Python keywords and when this IDL file is translated into Python the IDLC Python plugin has been modified to add a prefix "_" to the found reserved keywords.

We have had into account all the posible declaration using IDL: structs, modules, enums, typedefs...

We have tested broadly this change. The developed and executed tests can be review in detail in this git repository:

https://github.com/javiersorianoruiz/cyclonedds-python-checkReservedKeyword

These are not the formal tests that maybe are required by cyclonedds-python project, but they have allowed us to verify that the implementation is right and have no colateral issues.

Maybe, It would be needed to develop another test batery based on pytest and integrated into cyclonedds-python project, following your testing architecteture, What do you think?, Is it needed?

### DIFF
--- a/cyclonedds/idl/_xt_builder.py
+++ b/cyclonedds/idl/_xt_builder.py
@@ -974,9 +974,19 @@ class XTBuilder:
 
     @classmethod
     def _xt_complete_struct_member(cls, entity: Type[IdlStruct], name: str, _type: type) -> xt.CompleteStructMember:
+        #Reserved Python keywords support (Issue 105)
+        realName=name
+        if (name in get_idl_field_annotations(entity)):
+            if ("name" in get_idl_field_annotations(entity)[name]):
+                realName=get_idl_field_annotations(entity)[name]["name"]
+
+        #######################
+
         return xt.CompleteStructMember(
             common=cls._xt_common_struct_member(entity, name, _type, False),
-            detail=cls._xt_complete_member_detail(entity, name, _type)
+            #Reserved Python keywords support (Issue 105):
+            detail=cls._xt_complete_member_detail(entity, realName, _type)
+            #######################
         )
 
     @classmethod
--- a/cyclonedds/idl/_xt_builder.py
+++ b/cyclonedds/idl/_xt_builder.py
@@ -974,12 +974,9 @@ class XTBuilder:
 
     @classmethod
     def _xt_complete_struct_member(cls, entity: Type[IdlStruct], name: str, _type: type) -> xt.CompleteStructMember:
-        #Reserved Python keywords support (Issue 105)
-        realName=name
-        if (name in get_idl_field_annotations(entity)):
-            if ("name" in get_idl_field_annotations(entity)[name]):
-                realName=get_idl_field_annotations(entity)[name]["name"]
-
+        # Reserved Python keywords support (#105)
+        annotations = get_idl_field_annotations(entity).get(name)
+        realName = annotations["name"] if annotations and "name" in annotations else name
         #######################
 
         return xt.CompleteStructMember(
--- a/idlpy/src/context.c
+++ b/idlpy/src/context.c
@@ -36,6 +36,10 @@
 #include <sys/stat.h>
 #endif
 
+//Reserved Python keywords support (Issue 105)
+#include "types.h"
+/////////////
+
 /// * IDL context * ///
 typedef struct idlpy_module_ctx_s *idlpy_module_ctx;
 
@@ -530,7 +534,16 @@ static idl_retcode_t write_module_headers(FILE *fh, idlpy_ctx octx, idlpy_module
         idl_fprintf(fh, fmt_entities, entity_prefix, octx->idl_file);
 
         for(int i = 0; i < idlpy_ssos_size(ctx->this_idl_file->entities); ++i) {
-            idl_fprintf(fh, fmt_entity, i > 0 ? ", " : "", idlpy_ssos_at(ctx->this_idl_file->entities, i));
+
+           //Reserved Python keywords support (Issue 105)
+            const char *name = idlpy_ssos_at(ctx->this_idl_file->entities, i);
+            const char *nameModified=get_c_name(name);
+            if (nameModified)
+                name = nameModified;
+            
+            idl_fprintf(fh, fmt_entity, i > 0 ? ", " : "", name);
+            
+            ///////////////////////////
         }
         idl_fprintf(fh, "\n");
     }
@@ -551,7 +564,15 @@ static idl_retcode_t write_module_headers(FILE *fh, idlpy_ctx octx, idlpy_module
 
     if (idlpy_ssos_size(ctx->this_idl_file->entities) > 0) {
         for(int i = 0; i < idlpy_ssos_size(ctx->this_idl_file->entities); ++i) {
-            idl_fprintf(fh, "\"%s\", ", idlpy_ssos_at(ctx->this_idl_file->entities, i));
+
+            //Reserved Python keywords support (Issue 105)
+            const char *name = idlpy_ssos_at(ctx->this_idl_file->entities, i);
+            const char *nameModified=get_c_name(name);
+            if (nameModified)
+                name = nameModified;
+            
+            idl_fprintf(fh, "\"%s\", ", name);
+            ///////////////////////////
         }
     }
 
--- a/idlpy/src/context.c
+++ b/idlpy/src/context.c
@@ -37,7 +37,7 @@
 #endif
 
 //Reserved Python keywords support (Issue 105)
-#include "types.h"
+#include "naming.h"
 /////////////
 
 /// * IDL context * ///
@@ -536,13 +536,8 @@ static idl_retcode_t write_module_headers(FILE *fh, idlpy_ctx octx, idlpy_module
         for(int i = 0; i < idlpy_ssos_size(ctx->this_idl_file->entities); ++i) {
 
            //Reserved Python keywords support (Issue 105)
-            const char *name = idlpy_ssos_at(ctx->this_idl_file->entities, i);
-            const char *nameModified=get_c_name(name);
-            if (nameModified)
-                name = nameModified;
-            
+            const char *name = filter_python_keywords(idlpy_ssos_at(ctx->this_idl_file->entities, i));
             idl_fprintf(fh, fmt_entity, i > 0 ? ", " : "", name);
-            
             ///////////////////////////
         }
         idl_fprintf(fh, "\n");
@@ -566,11 +561,7 @@ static idl_retcode_t write_module_headers(FILE *fh, idlpy_ctx octx, idlpy_module
         for(int i = 0; i < idlpy_ssos_size(ctx->this_idl_file->entities); ++i) {
 
             //Reserved Python keywords support (Issue 105)
-            const char *name = idlpy_ssos_at(ctx->this_idl_file->entities, i);
-            const char *nameModified=get_c_name(name);
-            if (nameModified)
-                name = nameModified;
-            
+            const char *name = filter_python_keywords(idlpy_ssos_at(ctx->this_idl_file->entities, i));
             idl_fprintf(fh, "\"%s\", ", name);
             ///////////////////////////
         }
--- a/idlpy/src/types.c
+++ b/idlpy/src/types.c
@@ -25,7 +25,32 @@
 #include "naming.h"
 #include "types.h"
 
+//Reserved Python keywords support (Issue 105):
+static const char *python_keywords[] = {
+#define _py_(str) "_" str
+    _py_("False"), _py_("None"), _py_("True"), _py_("and"), _py_("as"), _py_("assert"), 
+    _py_("break"), _py_("class"), _py_("continue"), _py_("def"), _py_("del"), _py_("elif"), 
+    _py_("else"), _py_("except"), _py_("finally"), _py_("for"), _py_("from"), _py_("global"), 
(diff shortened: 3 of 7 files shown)


cyclonedds/idl/_xt_builder.py
```diff
@@ -974,9 +974,19 @@ def _xt_minimal_struct_member(cls, entity: Type[IdlStruct], name: str, _type: ty
 
     @classmethod
     def _xt_complete_struct_member(cls, entity: Type[IdlStruct], name: str, _type: type) -> xt.CompleteStructMember:
+        #Reserved Python keywords support (Issue 105)
+        realName=name
```
thijsmie, line 978: This needs a little bit of formatting and removing the duplicate dict access. I suggest the following:

````python
# Reserved Python keywords support (#105)
annotations = get_idl_field_annotations(entity).get(name)
real_name = annotations["name"] if annotations and "name" in annotations else name

File: idlpy/src/context.c [hunk 2]
```diff
@@ -530,7 +534,16 @@ static idl_retcode_t write_module_headers(FILE *fh, idlpy_ctx octx, idlpy_module
         idl_fprintf(fh, fmt_entities, entity_prefix, octx->idl_file);
 
         for(int i = 0; i < idlpy_ssos_size(ctx->this_idl_file->entities); ++i) {
-            idl_fprintf(fh, fmt_entity, i > 0 ? ", " : "", idlpy_ssos_at(ctx->this_idl_file->entities, i));
+
+           //Reserved Python keywords support (Issue 105)
+            const char *name = idlpy_ssos_at(ctx->this_idl_file->entities, i);
```
thijsmie, line 539: If you change `get_c_name` to return the name unmodified when it is not a keyword you can write:
```c
const char *name = get_c_name(idlpy_ssos_at(ctx->this_idl_file->entities, i));
```
However, I suspect the `ctx->this_idl_file->entities` SSOS should contain already filtered names, so here the call is unnecessary?

idlpy/src/types.c
```diff
@@ -25,7 +25,32 @@
...
+    _py_("False"), _py_("None"), _py_("True"), _py_("and"), _py_("as"), _py_("assert"), 
+    _py_("break"), _py_("class"), _py_("continue"), _py_("def"), _py_("del"), _py_("elif"), 
+    _py_("else"), _py_("except"), _py_("finally"), _py_("for"), _py_("from"), _py_("global"), 
+    _py_("if"), _py_("import"), _py_("in"), _py_("is"), _py_("lambda"), _py_("nonlocal"), 
+    _py_("not"), _py_("or"), _py_("pass"), _py_("raise"), _py_("return"), _py_("try"), 
+    _py_("while"), _py_("with"), _py_("yield")
+#undef _py_
+};
+
+const char *get_c_name(const char *name)
```
thijsmie, line 40: I think `get_c_name` is not very descriptive to what it actually does: escape python keywords. I think it is actually prudent to define `idlpy_identifier` that calls `idl_identifier` and then filters keywords. This simplifies your code in quite a few places since there is no modified vs non-modified handling needed.

There are a couple places where a name of something that is not a node is determined. For that you can define a `filter_python_keywords` function and use that inside `idlpy_identifier`.

There is a few non-keyword things you should also add to the filter, namely `idl`, `annotate`, `types`, `auto`, `TYPE_CHECKING`, `Optional` and `dataclass` since they are occuring as imports at the top of any idl python output. See also https://github.com/eclipse-cyclonedds/cyclonedds-python/issues/161.

idlpy/src/types.c
```diff
@@ -146,13 +179,21 @@ emit_field(
 
     char *type = typename(ctx, type_spec);
 
+    //printf("emit_field name : %s , type %s \n", name , type);
+
     if (idl_is_default_case(parent)) {
+
+        //printf("emit_field idl_is_default_case name : %s , type %s \n", name , type);
```
thijsmie, line 186: Please don't leave commented code in the PR.

idlpy/src/types.c
```diff
@@ -146,13 +179,21 @@ emit_field(
...
+
         char *ctype;
         idl_asprintf(&ctype, "types.default[%s]", type);
         free(type);
         type = ctype;
     }
     else if (idl_is_case(parent)) {
+
+        //printf("emit_field idl_is_case name : %s , type %s \n", name , type);
```
thijsmie, line 195: Please don't leave commented code in the PR.

idlpy/src/types.c
```diff
@@ -175,6 +216,9 @@ emit_field(
     }
 
     if (idl_is_member(parent)) {
+
+        //printf("emit_field idl_is_member name : %s , type %s \n", name , type);
```
thijsmie, line 220: Please don't leave commented code in the PR.

idlpy/src/types.c
```diff
@@ -185,42 +229,68 @@ emit_field(
...
-    idlpy_ctx_printf(ctx, "\n    %s: %s", name, type);
+    //Reserved Python keywords support (Issue 105):
+    const char *nameField = name;
+    const char *nameFieldModified=get_c_name(name);
+    if (nameFieldModified)
+        nameField =  nameFieldModified;
+    
+    idlpy_ctx_printf(ctx, "\n    %s: %s", nameField, type);
+
+    if (nameFieldModified) {
```
thijsmie, line 240: If you switch everything to `idlpy_identifier` this will become `if (name != idl_identifier(node))`. Note that `strcmp` is not needed, we're comparing constant pointers and that is fine in this case.

idlpy/src/types.c
```diff
@@ -185,42 +229,68 @@ emit_field(
...
+        idlpy_ctx_printf(ctx, "\n    annotate.member_name(\"%s\",\"%s\")", nameFieldModified,name);
+
+    }
 
     if (idl_is_member(parent)) {
+
+        //printf("emit_field idl_is_member2 name : %s , type %s \n", name , type);
```
thijsmie, line 247: Please don't leave commented code in the PR.

idlpy/src/types.c
```diff
@@ -699,8 +848,19 @@ emit_const(
 
     const idl_literal_t *literal = ((const idl_const_t *)node)->const_expr;
 
+    //printf("------- emit_const: name %s \n",idl_identifier(node));
```
thijsmie, line 851: Please don't leave commented code in the PR.

idlpy/src/types.h
```diff
@@ -17,4 +17,8 @@
 
 idl_retcode_t generate_types(const idl_pstate_t *pstate, idlpy_ctx ctx);
 
+//Reserved Python keywords support (Issue 105)
+const char *get_c_name(const char *name);
```
thijsmie, line 21: Maybe this function should live in `naming.h` instead?

thijsmie (changes requested):
Hi @javiersorianoruiz,

This looks already really nice! I have a couple things that need some attention, mostly stylistic. There is one feature you missed, we can also resolve types dynamically from the network using the `XTInterpreter`. The names in `XTInterpreter._make_complete_struct`, `XTInterpreter._make_complete_union` etc also need to take care with python keywords. I can also handle this myself in a followup PR, just let me know.

cyclonedds/idl/_xt_builder.py
javiersorianoruiz: Hello @thijsmie , thank you for your fast answer.

About "we can also resolve types dynamically from the network using the XTInterpreter", How do we have to do this? How do we have to write the IDL file for the method XTInterpreter._make_complete_struct or XTInterpreter._make_complete_union to be executed during the translation from IDL to Python ?

If you give us more information about how to reproduce this conversion, we can try to implement the change ourselves as we have done with the rest of cases.


About your suggestion "annotations = get_idl_field_annotations(entity).get(name) ...", you are right, your implementation is smarter, we will change it.

idlpy/src/context.c
javiersorianoruiz: OK, we can do as you suggest, but the cyclonedds-cxx implementation (regarding also using C++ reserved keywords) implements this function (file `idlpy/src/types.c`):

```
const char *get_c_name(const char *name)
{
  //printf("get_c_name: %s \n", name);
  static const size_t n = sizeof(python_keywords)/sizeof(python_keywords[0]);

  /* search through the Python keyword list */
  for (size_t i=0; i < n; i++) {
    if (strcmp(python_keywords[i] + 1, name) == 0)
      return python_keywords[i];
  }

  return NULL;
}
```

**In order to be consistent with other implementations (cxx versus Python), we have followed this example, so, what do you think we have to do at the end?**


Regarding to "_However, I suspect the ctx->this_idl_file->entities SSOS should contain already filtered names, so here the call is unnecessary?_"  

Yes indeed,  `ctx->this_idl_file->entities` does contain the filtered names, so here, the call is unnecessary, in one of our tests, we have worked with this IDL file:

```
module pass {
    struct struct_test {
        char var;
    };
};
```

Our version of the plugin Python IDLC compiler, generates a folder con name "_pass" and inside the files:

```
__init__.py

from ._test import struct_test
__all__ = ["struct_test", ]
```
```
_test.py
from enum import auto
from typing import TYPE_CHECKING, Optional

import cyclonedds.idl as idl
import cyclonedds.idl.annotations as annotate
import cyclonedds.idl.types as types

# root module import for resolving types
import _pass


@dataclass
@annotate.final
class struct_test(idl.IdlStruct, typename="_pass.struct_test"):
    var: types.char
```

So, our version is working properly, and we have tested it with a publisher and subscriber in order to check that a message with this format can be sent and received.

idlpy/src/types.c
javiersorianoruiz: Ok, so, we understand you don't want us to follow the cxx way of implementation. 

To confirm with you:

1. Modify name of function `get_c_name `to `filter_python_keywords `
2. `filter_python_keywords` never returns NULL, returns the original name or the original name with the prefix "_".
3. We implement a new function `const char *idlpy_identifier(const void *node)` which internally is going to call:

```
   name = idl_identifier(node) 
   name = filter_python_keywords(name)
   return  name;
```

About more keywords, OK, we will also review this #161 to add all the required keywords. We have filtered the Python keywords indicated in this web page:

https://www.w3schools.com/python/python_ref_keywords.asp
https://www.geeksforgeeks.org/python-keywords/

But, only for curiosity, why do we have to add other keywords that are not considered reserved?, We suppose that the Python code generated doesn't work if it contains any of the "extra" keywords, does it?

idlpy/src/types.c
javiersorianoruiz: ha ha ha, it was an oversight, we have already deleted this printf and another one that we have also found

idlpy/src/types.c
thijsmie: The proposed modifications sound okay to me.

As for your questions, indeed, they are not keywords but are used in the generated code:

```python
from enum import auto
from typing import TYPE_CHECKING, Optional

import cyclonedds.idl as idl
import cyclonedds.idl.annotations as annotate
import cyclonedds.idl.types as types
```

idlpy/src/types.c
javiersorianoruiz: all the printfs that we have added have been deleted

idlpy/src/context.c
thijsmie: There is indeed no need to follow the C++ code here. We're going for maximum readability always, if it is used differently in one other place that doesn't really matter to me.

idlpy/src/types.c
javiersorianoruiz: all the printfs that we have added have been deleted

idlpy/src/types.c
Bit-Seq: After reviewing it today, all the printfs we have added have been deleted.

idlpy/src/types.c
Bit-Seq: After reviewing it today, all the printfs we have added have been deleted.

idlpy/src/types.h
Bit-Seq: Ok. That header seems more proper to contain the method. We will move it.

Bit-Seq:
Another detail not mentioned above. True and False keywords are not supported, due to they are also keywords in the IDL.

The IDLC compiler returns a syntax error even before trying to convert the IDL file to Python, so IDLC does not support these words in an IDL file for any supported language.

Anyway, True and False are keywords in almost all languages, and it is rare a lot to try to define an object with those identifiers.

javiersorianoruiz:
We have updated the code with your suggested changes. So you can advance in the pull request and merge process.

Everything is working properly in our testing: https://github.com/javiersorianoruiz/cyclonedds-python-checkReservedKeyword

We have pending:

```
About "we can also resolve types dynamically from the network using the XTInterpreter", How do we have to do this? How do we have to write the IDL file for the method XTInterpreter._make_complete_struct or XTInterpreter._make_complete_union to be executed during the translation from IDL to Python?

If you give us more information about how to reproduce this conversion, we can try to implement the change ourselves as we have done with the rest of cases.
``` 

We have added also the keywords you ask for them, so, consider closing Issue #161  if the pull request is ok for you.

thijsmie (approved):
I think it is okay like this! Thank you so much for the contribution. The last remaining point I mentioned about the network types is something I'll tackle later.

PR #164, opened 2022-11-11
Fix struct inheritance

### LINKED ISSUE #162: Inheritance in cyclonedds-python not working or not implemented
Independent structs and structs inheritance seem to generate the same python code. Let’s have an example.
 
Example of two different structs in an IDL file:
 
```
module module_test {
    struct struct_test_A {
        char var;
    };
    struct struct_test_B {
        char varr;
    };
};
```
 
 Example of structs inheritance in another IDL file:
 
```
module module_test {
    struct struct_test_A {
        char var;
    };
    struct struct_test_B : struct_test_A {
        char varr;
    };
};
```
 
When we (@Bit-Seq,@javiersorianoruiz) translate from IDL to Python the examples above, both give the same result:

 
```

from enum import auto
from typing import TYPE_CHECKING, Optional

import cyclonedds.idl as idl
import cyclonedds.idl.annotations as annotate
import cyclonedds.idl.types as types

# root module import for resolving types
import module_test

@dataclass
@annotate.final
class struct_test_A(idl.IdlStruct, typename="module_test.struct_test_A"):
    var: types.char

@dataclass
@annotate.final
class struct_test_B(idl.IdlStruct, typename="module_test.struct_test_B"):
    varr: types.char
```

There seems to be a bug and the inheritance case is not implemented in Python.
Why different IDL files give the same result?

thijsmie:
This fixes #162 and some related issues around struct inheritance. @jordialexalar will you please test if this fixes your issue?

### DIFF
--- a/cyclonedds/idl/_main.py
+++ b/cyclonedds/idl/_main.py
@@ -357,6 +357,12 @@ class IdlMeta(type):
         if typename:
             namespace["__idl_typename__"] = typename
 
+        for base in __bases:
+            if isinstance(base, IdlMeta):
+                namespace["__annotations__"].update(base.__annotations__)
+                namespace["__idl_annotations__"].update(base.__idl_annotations__)
+                namespace["__idl_field_annotations__"].update(base.__idl_field_annotations__)
+
         IDLNamespaceScope.enter(namespace)
         return namespace
 
--- a/cyclonedds/idl/_type_normalize.py
+++ b/cyclonedds/idl/_type_normalize.py
@@ -85,7 +85,7 @@ def _make_extended_type_hints(cls: Any) -> Dict[str, Any]:
 
 
 def _normalize_idl_class(cls: Any) -> None:
-    if hasattr(cls, "__idl_type_annotations__"):
+    if "__idl_type_annotations__" in cls.__dict__:
         return
     if cls.__name__.startswith("Idl") and cls.__module__ == "cyclonedds.idl":
         return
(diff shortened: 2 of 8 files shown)


jordialexalar:
Hi Thijs,

Thanks for your answer.
After testing the changes you made, I can say that inheritance works
properly, like for example in your file tests/test_struct_inheritance.py
<https://github.com/eclipse-cyclonedds/cyclonedds-python/pull/164/files#diff-bdd22c79cf592e46ebb88c7aabe75bc97f7eff3162a06f48bcbf764e5d05d890>
but
when we use inheritance in our examples of conexion , with file
publisher.py and subscriber.py,
another error arises.
If attached in this e-mail the files of the example and another word file
with the error that appears when the subscriber.py file is executed.
Perhaps you can help us with the problem.

Thanks for all

Jorge Alejandre

thijsmie:
Hi @jordialexalar,

Could you send the files you mentioned here or to my email opensource _at_ tmiedema.com, attaching them to the comment email doesn't seem to have worked.

thijsmie:
Hi @jordialexalar,

Thank you for your reproducer, it helped me track down a subtle way it could still go wrong. With this extra one-line change your example now works for me.

jordialexalar:
Hi Thijs,

Thanks for your answer.

After testing the changes you have made, I've just tried them,and as you
say now it works fine.

Thanks for all

Jorge Alejandre

PR #167, opened 2022-11-17
Update to tools.rst

PeterCadogan:
The title has changed from "Tools" to "The cyclonedds command line tool". Maybe we should rename the file to match the title.

I've added an introductory sentence and added a list of the subcommands, so that the reader can see whether this is the topic they are looking for without having to scan the entire page.

I've edited the text so that it is consistent.

### DIFF
--- a/docs/manual/tools.rst
+++ b/docs/manual/tools.rst
@@ -1,12 +1,21 @@
-Tools
-=====
-
-When you install the ``cyclonedds`` Python package you also get a ``cyclonedds`` command line tool with several subcommands.
-
 The ``cyclonedds`` command line tool
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+====================================
 
- The main help screen shows the commands available:
+Included with the ``cyclonedds`` Python package is the ``cyclonedds`` command line tool. 
+
+The command line tool provides a quick insight into what your DDS dataspace currently looks like. It allows you to verify that the public view of your applications matches what you had intended.
+
+The following is a list of the ``cyclonedds`` subcommands:
+
+* ``cyclonedds --help``: Shows the help for the ``cyclonedds`` subcommands.
+* ``cyclonedds ls``: Shows the entities and their QoS settings in your DDS system.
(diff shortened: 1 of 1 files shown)


docs/manual/tools.rst
```diff
@@ -1,47 +1,62 @@
...
 ``cyclonedds ls``
 -----------------
 
+The ``ls`` subcommand shows the entities and their QoS settings in your DDS system.
+
 .. image:: static/images/cyclonedds-ls-help.svg
     :alt: ``cyclonedds ls --help``
 
-The ``ls`` subcommand shows you the entities in your DDS system and their QoS settings. For example, here is the output when running the ``Vehicle`` example from this repo in the background:
+The following example screen shows the output from the ``Vehicle`` example when running this repo in the background:
```
thijsmie, line 31: The reword here doesn't quite match what is intended. Maybe 
```
The following screen shows the output from ``cyclonedds ls`` while running the Python ``Vehicle`` example in the background:
```

docs/manual/tools.rst
```diff
@@ -1,47 +1,62 @@
...
 ``cyclonedds ps``
 -----------------
 
+The ``ps`` subcommand lists the applications in your DDS system. Note: This depends on 'Participant Properties', which are known as "QoS properties" for DDS participants (merged into CycloneDDS version 0.10.0).
+
 .. image:: static/images/cyclonedds-ps-help.svg
     :alt: ``cyclonedds ps --help``
 
-The ``ps`` subcommand shows you the applications in your DDS system. Note that this depends on so called 'Participant Properties', tactfully named QoS properties in DDS participants. These were merged into CycloneDDS for version 0.10.0. Here is an example of the output when running the ``Vehicle`` example from this repo in the background on a single host:
+The following example screen shows the output from the ``Vehicle`` example when running this repo in the background on a single host:
```
thijsmie, line 44: Same wording problem as previous comment.

thijsmie (changes requested):
Thanks for your contributions! I have two small things where the rewording doesn't quite work out, for the rest it looks good!

thijsmie (approved):

PR #176, opened 2022-12-09
Fix CI

thijsmie:
Somewhere the CI for Python got really broken and it took me some time to get it running properly again. There is a very curious test failure in CI for Windows now that might be indicative of an actual bug, fixing it is out of scope for this PR, I'll look into it further.

### DIFF
--- a/.azure/templates/build-cyclone.yml
+++ b/.azure/templates/build-cyclone.yml
@@ -15,53 +15,51 @@
 #
 
 steps:
+  - task: UsePythonVersion@0
+  - bash: |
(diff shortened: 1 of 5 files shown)


poetinger:
I wonder why Python 3.10 takes 20 minutes longer for its Fuzz tests, compared with Python 3.7-3.9.

poetinger (approved):

thijsmie:
The time difference has nothing to do with the python version, it is caused by the luck of the draw when it comes to random types. 😇

PR #177, opened 2022-12-16
Update image paths in README.

thijsmie:


NoxPardalis (approved):
We should definitely merge this change in so the README looks as good as it used to. There's no way the CI failures are a result of these changes.

Issue #187, opened 2023-03-17
Support for default domains other than 0


gmartin82:
At present the Cyclone DDS Python binding defaults to using 0 as the default domain when creating domain participants. Further, no option is present in the binding to allow users to automatically create domain participants on the default domain as determined by the core Cyclone DDS implementation and the configuration files.

It would be useful for the Python binding to provide a mechanism which allows users to create domain participants on the default domain as determined by the Cyclone DDS XML configuration file specified through the CYCLONEDDS_URI environment variable.

PR #188, opened 2023-03-17
#187 Modified implementation to construct domain participants on the default domain

gmartin82:
Modified the implementation to construct domain participants on the default domain (as determined by the Cyclone DDS configuration) by default unless users provide a domain Id when constructing a domain participant.

The proposed solution could potentially affect existing applications as, depending on configuration, user applications may run by default on a domain other than 0 as they would have previously. However, this may actually be desirable. An alternative if this is problematic would be to add a classmethod to Domain to return the default domain Id constant in a similar way to the C++ binding.

I've also updated the examples and tooling to use the default domain by default.

I also found the async example wasn't working with the regular vehicle publisher. I've modified the implementation to ensure the examples are able to communicate as part of the testing.

### DIFF
--- a/cyclonedds/tools/cli/data/__init__.py
+++ b/cyclonedds/tools/cli/data/__init__.py
@@ -1,5 +1,5 @@
 import time
-from typing import Any
+from typing import Any, Optional
 from cyclonedds import qos, domain, sub, topic
 
 from ..utils import LiveData
@@ -7,13 +7,16 @@ from ..utils import LiveData
 
 def subscribe(
     live: LiveData,
-    domain_id: int,
+    domain_id: Optional[int],
     topic_name: str,
     datatype: Any,
     topicqos: qos.Qos,
     readerqos: qos.Qos,
 ):
-    dp = domain.DomainParticipant(domain_id)
+    if domain_id is None:
+        dp = domain.DomainParticipant()
+    else:
+        dp = domain.DomainParticipant(domain_id)
     tp = topic.Topic(dp, topic_name, datatype, qos=topicqos)
     rd = sub.DataReader(dp, tp, qos=readerqos)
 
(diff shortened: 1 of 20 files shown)


cyclonedds/tools/cli/data/__init__.py
```diff
@@ -1,19 +1,22 @@
...
     live: LiveData,
-    domain_id: int,
+    domain_id: Optional[int],
     topic_name: str,
     datatype: Any,
     topicqos: qos.Qos,
     readerqos: qos.Qos,
 ):
-    dp = domain.DomainParticipant(domain_id)
+    if domain_id is None:
```
eboasson, line 16: Pardon my ignorance, but why can't you pass `domain_id` directly to `domain.DomainParticipant`?

gmartin82: The DomainParticipant constructor was originally written with the expectation that users would either pass an int for the domain_id parameter or omit the parameter which would then cause the domain_id to default to a value of 0. My intention was to change this so it defaulted to DDS_DOMAIN_DEFAULT instead but looking back at the code I've made a mistake with the type hinting as the Optional hint allows None to be passed which wasn't the intention. I've corrected the type hinting for the constructor back to its original form which conveys this intent.

In the CLI tooling, however, I did want the functions to accept a domain_id of None which was intended to indicate that the domain Id argument had not been specified at the command line. Passing this None value through to the DomainParticipant constructor would actually result in the domain_id parameter being set to None instead of the default value which is not desired.

Also on reflection, I think these if/else DomainParticipant creation statements can be made more concise through the use of ternary statements such as:

dp = domain.DomainParticipant(domain_id) if domain_id else domain.DomainParticiant()

If you prefer this syntax I can update the PR.

gmartin82: I should also add that Click (the library used for handling the CLI arguments) passes an argument of None by default to the handling function when users imit the option.

gmartin82: @eboasson have you had a chance to take another look at this?

eboasson (approved):
Thanks @gmartin82 for pinging me! I had completely overlooked the notification of your answering my question (or I saw them at an inconvenient time and then forgot about them). So yes, I see why you chose this approach. This is fine with me.

PR #192, opened 2023-05-18
Fix checking if value is None in _serialize.py

morkovka1337:
This patch makes check more pythonic way and fixes an error of checking array values


thijsmie:
Hi @morkovka1337,

According to the type this value can already be not be None (use `Optional` for that). This if statement here is purely to guard against zero-length-arrays (which evaluate to `False` when forced to boolean). The pre-PR state of this code was correct :)

eboasson:
Trusting @thijsmie here, and so closing issue. But I also want to thank you @morkovka1337. Most grateful to everyone who puts effort into helping develop Cyclone!

PR #195, opened 2023-06-27
Fix fully qualified symbol name outside a module

eboasson:
The code in `absolute_name` constructed the correct name, then patched it to get the module prefix for updating the list of referenced modules, then backed out the patch to return the full name.

The patching handled symbols outside any module incorrectly, resulting in weird references of names ending in a single quote by patching the opening quote instead of the (non-existent) separating dot, and names starting with a dot and ending in a single quote because the patch was undone by unconditionally putting a dot in where there had been none.

This makes the entire module prefix and module referencing conditional upon the existence of an enclosing module and furthermore documents/asserts the structure of the generated name.

### DIFF
--- a/idlpy/src/naming.c
+++ b/idlpy/src/naming.c
@@ -233,8 +233,9 @@ char* typename_unwrap_typedef(idlpy_ctx ctx, const void *node)
     }
 }
 
-char *absolute_name(idlpy_ctx ctx, const void *node)
+char *absolute_name(idlpy_ctx ctx, const void *vnode)
 {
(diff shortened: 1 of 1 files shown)


poetinger (approved):
Thank you for the detailed explanation behind the bug and your solution. LGTM

eboasson:
Thanks @poetinger for reviewing this! I can't quite see how the test failures could have been caused by this. Even though I feel less comfortable making such a call for the python binding than I do for failures in the `cyclonedds` repository, I am merging it. Hopefully I'm making the right call ...

PR #233, opened 2024-02-22
Add hash generation to pysertype.c, add OOM test.  Closes #232

### LINKED ISSUE #232: OOM when using a QoS with `max_instances` larger than a small value (or the default: -1)
The following reproducer results in an OOM on x64-linux, Debian 11.8.  This is running inside a Docker container.  Host machine has 32GB RAM.

I'm using CycloneDDS commit: `2cfdc1b7cfa1c3d9b37b09109113b1f5d3b01802:  05 Feb 24`
and CycloneDDS Python commit: `637cfe583a5078af41c481c98daf7474fdb2a786:  06 Sep 23`
and installing the Python library via `CYCLONEDDS_HOME=${CYCLONEDDS_HOME} python3 -m pip install /path/to/cyclonedds-python`

    from cyclonedds.pub import DataWriter
    from cyclonedds.sub import DataReader
    from cyclonedds.topic import Topic
    from cyclonedds.domain import DomainParticipant
    from cyclonedds.idl.annotations import key as _key
    from cyclonedds.idl import IdlStruct
    from cyclonedds.qos import Policy, Qos

    @dataclass
    class KeyedString(IdlStruct, typename="DDS.KeyedString"):
        "Represents a key-value string pair"
        key: str
        value: str
        _key("key")

    def test_qos(qos_to_test: Qos):

        N_ITERATIONS = 50

        print(f"Testing QoS: {str(qos_to_test)}")

        # create DDS entities
        participant = DomainParticipant()
        topic = Topic(domain_participant=participant,
                    topic_name="foo", data_type=KeyedString)
        writer = DataWriter(publisher_or_participant=participant,
                            topic=topic, qos=qos_to_test)
        _ = DataReader(subscriber_or_participant=participant,
                    topic=topic, qos=qos_to_test)

        # test N samples
        sample = KeyedString(key="Hello", value="World")
        for i in range(N_ITERATIONS):
            print(f"Sample {i}")
            sample.value = str(i)
            writer.write(sample=sample)
            writer.dispose(sample=sample)

        # test N instances
        for i in range(N_ITERATIONS):
            print(f"Instance {i}")
            instance = KeyedString(key=str(i), value=str(i))
            writer.write(sample=instance)
            writer.dispose(sample=instance)

    if __name__ == '__main__':

        # test using a QoS with lower instance limits
        test_qos(qos_to_test=Qos(Policy.ResourceLimits(max_instances=10)))

        # test using a QoS with higher instance limits
        test_qos(qos_to_test=Qos(Policy.ResourceLimits(max_instances=50)))

Output:

    Testing QoS: Qos(Policy.ResourceLimits(max_samples=-1, max_instances=10, max_samples_per_instance=-1))
    Sample 0
    Sample 1
    Sample 2
    ...
    Sample 48
    Sample 49
    Instance 0
    Instance 1
    Instance 2
    ...
    Instance 48
    Instance 49
    Testing QoS: Qos(Policy.ResourceLimits(max_samples=-1, max_instances=50, max_samples_per_instance=-1))
    Sample 0
    Sample 1
    Sample 2
    ...
    Sample 48
    Sample 49
    Instance 0
    Instance 1
    Instance 2
    ...
    Instance 30
    Instance 31
    Killed

Running `apt-get install time; /usr/bin/time -v python3 path/to/my/test.py` results in `Command terminated by signal 9...Maximum resident set size (kbytes): 14533312`
It doesn't make a difference if I call `dispose` or not on the publisher side
It doesn't make a difference if I add a sleep(1.0) between write iterations
It does not seem related to the message size, as I first encountered this error at a similar iteration count using much larger messages than `KeyedString`
`take`ing from the reader within each iteration has no effect
Adding QoS properties of `ReaderDataLifecycle(autopurge...=duration(nanoseconds=1)), Policy.History.KeepLast(1),
        Policy.WriterDataLifecycle(autodispose=True)` has no effect
Registering/unregistering each instance has no effect

This appears to be a bug in the implementation.  Is there a workaround other then setting `max_instances` to some arbitrary limit?

clunietp:
Added sample/instance hash value generation.  Without a hash value, users will receive an OOM when using >= 32 instances within a single datareader.

### DIFF
--- a/clayer/pysertype.c
+++ b/clayer/pysertype.c
@@ -127,17 +127,6 @@ static ddspy_serdata_t *ddspy_serdata_new(const struct ddsi_sertype* type, enum 
     return new;
 }
 
-static void ddspy_serdata_populate_hash(ddspy_serdata_t* this, ddsi_sertype_t* type) {    
-    ddsi_serdata_t* sd = cserdata(this);
-    ddsi_keyhash_t kh;
-    type->serdata_ops->get_keyhash(sd,&kh,true);
-    // consume the first N bytes of the MD5
-    memcpy(&(sd->hash), kh.value, sizeof(sd->hash));
-    // xor with the basehash of the type
-    sd->hash ^= type->serdata_basehash;
-    assert(sd->hash);
-}
-
 static void ddspy_serdata_populate_key(ddspy_serdata_t* this)
 {
     if (csertype(this)->keyless) {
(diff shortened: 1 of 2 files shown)


eboasson (approved):
Thank you very much for this one, too, @clunietp ☺️

PR #236, opened 2024-03-11
populate_key must differentiate between key/data

eboasson:
If the input to ddspy_serdata_populate_key is actually a key, then one must not try to interpret it as a full sample.


dpotman (approved):
LGTM!

PR #237, opened 2024-03-15
Adding topic inheritance test

adrianomarto:
Tests if derived topics can be published and subscribed.
This test aims to demonstrate the issue #238.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,6 +147,7 @@ setup(
     extras_require={
         "dev": [
             "pytest>=6.2",
+            "pytest-asyncio",
             "pytest-cov",
             "pytest-mock",
             "flake8",
--- a/tests/test_topic_inheritance.py
+++ b/tests/test_topic_inheritance.py
@@ -1,0 +1,111 @@
+import asyncio
+import pytest
+
+import cyclonedds.idl as idl
+import cyclonedds.idl.annotations as annotate
+
+from dataclasses import dataclass
+from cyclonedds.domain import DomainParticipant
+from cyclonedds.topic import Topic
+from cyclonedds.pub import DataWriter
+from cyclonedds.sub import DataReader
+from cyclonedds.util import duration
+
+
+@dataclass
+@annotate.mutable
+@annotate.autoid("sequential")
+class Base(idl.IdlStruct, typename="Hierarchy.Base"):
+    fieldA: str
+
+
+@dataclass
+@annotate.mutable
+@annotate.autoid("sequential")
+class Derived(Base, typename="Hierarchy.Derived"):
+    fieldB: str
+
+
+@pytest.mark.asyncio
+async def test_base_topic():
+    '''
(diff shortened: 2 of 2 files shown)


eboasson:
@adrianomarto I just happened to read this scary pair of sentences in a GitHub email:

> This test aims to demonstrate what I believe to be a bug. I still need to figure out how to fix it.
>
> A domain participant can only publish or subscribe to a derived topic if it also publishes or subscribes to its base topic.

So it turns out you've edited it since, but I looked at the test code anyway. It looks to me like the second `assert` in `test_derived_topic` is inconsistent with the commenting out of the base class case. Anyway, I'm sure you've seen this yourself already as well.

The other bit is that with it then surprised me that the CI is green. It turns out it skips the tests because of the `asyncio` bit. (It beats me why a test that is skipped like this is considered a passing test ...) I don't know the Python ecosystem, but with a bit of luck you know what needs to be done to solve this:
```
#[warning]async def functions are not natively supported and have been skipped.
You need to install a suitable plugin for your async framework, for example:
  - anyio
  - pytest-asyncio
  - pytest-tornasync
  - pytest-trio
  - pytest-twisted

test_topic_inheritance.py::test_derived_topic SKIPPED (async def
##[warning]async def functions are not natively supported and have been skipped.
You need to install a suitable plugin for your async framework, for example:
  - anyio
  - pytest-asyncio
  - pytest-tornasync
  - pytest-trio
  - pytest-twisted
```

In any case, you should be able to use the derived type for a topic without also using the base type as a topic. If that doesn't work, it's a bug, possibly a python-only bug.

adrianomarto:
@eboasson, I'd like to point out that you understood my intent correctly. I observed that behaviour while using Cyclonedds v0.10.4 and Cyclonedds-Python v0.10.2. I am still trying to figure out how to reproduce the same behaviour in the CI environment. Yes, I first thought that it was working correctly, but then I noticed that my test was being skipped

adrianomarto:
I am afraid we have a bug indeed. The test works fine when I publish and subscribe to the Base topic. It also works as expected when I publish and subscribe to the Base and Derived topics. However, the test fails when I only publish and subscribe to the Derived topic.

eboasson (commented):
Looks good except for one thing: in my understanding of the way all the Python setup happens, it also needs a reference to "pytest-asyncio" in [setup.py](https://github.com/eclipse-cyclonedds/cyclonedds-python/blob/a60ed7120eb488898dfda55a4827d000da1a8718/setup.py#L148), because otherwise a `pip3 install '.[dev]'` won't install it automatically if it isn't present yet.

eboasson (approved):
LGTM, thank you @adrianomarto !

CI build re-done after merging #239 looks good: https://dev.azure.com/eclipse-cyclonedds/cyclonedds-python/_build/results?buildId=5763&view=results

PR #239, opened 2024-03-18
Define (and correctly reference) the base type

### LINKED ISSUE #238: Cannot publish/subscribe to a derived topic without publishing/subscribing to its base topic
Publishing and subscribing to a derived topic does not work. I first observed that behaviour while using Cyclonedds v0.10.4 and Cyclonedds-Python v0.10.2 (plus the changes in the pull request #164). In that instance, the subscriber did not receive updates on the derived topic.

I created pull request #237 to demonstrate the issue. The behaviour is different in the CI environment (perhaps because it is based on the current master branch). The exception DDSException is raised in the CI environment when trying to instantiate the topic.

eboasson:
This includes a struct's base type in the set of types that are referenced by it, and it makes the complete type object use the complete type id (rather than the minimal one) to reference it.

Fixes #238


adrianomarto:
I performed some quick tests, and I confirm this fixes the issue: https://github.com/eclipse-cyclonedds/cyclonedds-python/issues/238. @eboasson , thanks for that!

Issue #241, opened 2024-04-01
The command "cyclonedds typeof" fails to reconstruct the IDL of a derived topic


adrianomarto:
I got a crash whilst running the command "cyclonedds typeof Derived".

Here is the IDL:

module Hierarchy
{
  @mutable struct Base
  {
    string fieldA;
  };

  struct Derived : Base
  {
    string fieldB;
  };
};

Here is the stack-trace:

Exception in thread Thread-1 (type_discovery):
Traceback (most recent call last):
  File "/usr/lib64/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib64/python3.11/site-packages/cyclonedds/tools/cli/discovery/main.py", line 281, in type_discovery
    datatype, _ = dynamic.get_types_for_typeid(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/cyclonedds/dynamic.py", line 69, in get_types_for_typeid
    return XTInterpreter.xt_to_class(type_id, typemap)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/cyclonedds/idl/_xt_builder.py", line 1458, in xt_to_class
    main_type = cls._from_typeid(ident, state)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/cyclonedds/idl/_xt_builder.py", line 1503, in _from_typeid
    return cls._from_typeobject(ident, state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/cyclonedds/idl/_xt_builder.py", line 1524, in _from_typeobject
    return cls._make_complete_struct(ident, complete_obj.struct_type, state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/cyclonedds/idl/_xt_builder.py", line 1592, in _make_complete_struct
    struct = make_idl_struct(
[...]

PR #242, opened 2024-04-01
Testing the command "cyclonedds typeof" with a derived topic

adrianomarto:
The added test currently fails due to the issue #241.

### DIFF
--- a/tests/test_topic_inheritance.py
+++ b/tests/test_topic_inheritance.py
@@ -1,0 +1,111 @@
+import asyncio
+import pytest
+
+import cyclonedds.idl as idl
+import cyclonedds.idl.annotations as annotate
+
+from dataclasses import dataclass
+from cyclonedds.domain import DomainParticipant
+from cyclonedds.topic import Topic
+from cyclonedds.pub import DataWriter
+from cyclonedds.sub import DataReader
+from cyclonedds.util import duration
+
+
+@dataclass
+@annotate.mutable
+@annotate.autoid("sequential")
+class Base(idl.IdlStruct, typename="Hierarchy.Base"):
+    fieldA: str
+
+
+@dataclass
+@annotate.mutable
+@annotate.autoid("sequential")
+class Derived(Base, typename="Hierarchy.Derived"):
+    fieldB: str
+
+
+@pytest.mark.asyncio
+async def test_base_topic():
+    '''
+    Creates a publisher and a subscriber of the Base topic and checks if the
+    update sent by the publisher is received by the subscriber.
+    '''
+    base = Base(fieldA='Lorem')
+    tasks = [
+        _subscriber(Base),
+        _publisher(Base, base),
+    ]
+    results = await asyncio.gather(*tasks)
+    assert results[0] == results[1]
+
+
+@pytest.mark.asyncio
+async def test_derived_topic():
+    '''
+    Creates a publisher and a subscriber of the Derived topic and checks if the
+    update sent by the publisher is received by the subscriber.
+    '''
+    derived = Derived(
+        fieldA='Ipsum',
+        fieldB='Dolor')
+
+    tasks = [
+        _subscriber(Derived),
+        _publisher(Derived, derived),
+    ]
+
+    results = await asyncio.gather(*tasks)
+    assert results[0] == results[1]
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


tests/test_topic_inheritance.py
```diff
@@ -11,14 +11,15 @@
...
-
+BASE_IDL = 'module Hierarchy { @mutable struct Base { string fieldA; }; };'
 @dataclass
 @annotate.mutable
 @annotate.autoid("sequential")
 class Base(idl.IdlStruct, typename="Hierarchy.Base"):
     fieldA: str
 
 
+DERIVED_IDL = 'module Hierarchy { struct Derived : Base { string fieldB; }; };'
```
eboasson, line 22:
```suggestion
DERIVED_IDL = 'module Hierarchy { @mutable struct Base { string fieldA; }; @mutable struct Derived : Base { string fieldB; }; };'
```

eboasson (commented):
I may have fixed the problem of `cyclonedds typeof` on derived types, but I forgot to re-run the CI on this PR after that fix.

It turns out there are still some minor issues ...

For Python ≥ 3.8, it fails because the output of `typeof` doesn't match the expectation. The output now correctly includes the definition of `Base` and so I think the expectation needs to be adjusted.

On the CI, it fails for Python 3.7 with:
```
/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/subprocess.py:512: CalledProcessError
----------------------------- Captured stderr call -----------------------------
Traceback (most recent call last):
  File "/home/vsts/.local/bin/cyclonedds", line 5, in <module>
    from cyclonedds.tools.cli.main import cli
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/main.py", line 4, in <module>
    from .ls import ls
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/ls.py", line 6, in <module>
    from .discovery.main import ls_discovery
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/discovery/main.py", line 10, in <module>
    from ..idl import IdlType
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/idl.py", line 130
    return _type.length, *inner
                         ^
SyntaxError: invalid syntax
Traceback (most recent call last):
  File "/home/vsts/.local/bin/cyclonedds", line 5, in <module>
    from cyclonedds.tools.cli.main import cli
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/main.py", line 4, in <module>
    from .ls import ls
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/ls.py", line 6, in <module>
    from .discovery.main import ls_discovery
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/discovery/main.py", line 10, in <module>
    from ..idl import IdlType
  File "/home/vsts/.local/lib/python3.7/site-packages/cyclonedds/tools/cli/idl.py", line 130
    return _type.length, *inner
                         ^
SyntaxError: invalid syntax
----- generated Nunit xml file: /home/vsts/work/1/s/tests/test-output.xml ------
```
It seems to me that this an issue with the `typeof` implementation on Python 3.7. That's not a problem with this PR, but I do wonder if we shouldn't simply retire Python 3.7 support instead of trying to exclude this test in that one case.

You almost certainly have a better idea how relevant Python 3.7 still is. I wouldn't be surprised if it is the standard version in Ubuntu 20.04 or so, but if that's the case, then perhaps we should bump the Ubuntu builds to 22.04 just like we did for the Cyclone core library.

I would be grateful if you would share your view on this.

tests/test_topic_inheritance.py
adrianomarto: Hi @eboasson! Thanks for fixing the "typeof" command and circling back to this pull request. Your code change suggestion is correct, but I found another minor problem in the _type_checker function. I intend to handle the code changes when I return to the office in three weeks.

Stopping supporting Python 3.7 makes sense, as it has already reached its end-of-life. I have yet to check the Python version on Ubuntu or the end-of-life dates for its versions, but it would make sense to follow what has been done in CycloneDDS core. I would need to think about that a bit further to give you a better answer.

PR #246, opened 2024-04-24
Prepend _ to value, discriminator in union

### LINKED ISSUE #245: Setting Unions with no value
This is more of a cry for help than an issue. I am attempting to work with the OARIS IDL from the OMG. This spec uses a number of Union types to indicate if data is present, but for the life of me I cannot figure out how to use the union types when data is not present!

For example
cartesian_position_z_coordinate_type(discriminator=cartesian_coordinate_type, value=7)
Seems to parse ok, but
cartesian_position_z_coordinate_type()
cartesian_position_z_coordinate_type(discriminator=False)
cartesian_position_z_coordinate_type(discriminator=None)
Etc do not. 

I suspect I have the wrong end of the stick on unions and would appreciate a pointer!

eboasson:
The use of members named "value" and "discriminator" in a union are problematic because of the existence of properties on the IdlUnion class with these names.  (The error message are quite confusing.)  For example, this:

    union U switch (boolean) {
        case TRUE: string value;
    };

would result in a union where you could not set the value.  You might be able to get it, more or less by accident.  If instead of "value" the member would be named "discriminator", it'd be even worse.

The IDL spec has a mechanism for dealing with these problems: the use of a leading underscore as an keyword escape mechanism, both on the IDL and the language binding.  The Python already used this for keywords, now it also uses it for these two identifiers in unions.  Thus, the above now maps to:

    class U(idl.IdlUnion, discriminator=bool, discriminator_is_key=False, typename="U"):
        _value: types.case[[True], str]
        annotate.member_name("_value","value")

Trying to define a union with these members (without the leading underscore) now raises an exception.  The underscore is not in the TypeObjects and also stripped in the IDL output by `cyclonedds typeof`.

Fixes #245 (@FirebarSim, perhaps you can verify this claim?)

### DIFF
--- a/cyclonedds/idl/_main.py
+++ b/cyclonedds/idl/_main.py
@@ -472,6 +472,8 @@ class IdlUnionMeta(IdlMeta):
         for name, _type in new_cls.__annotations__.items():
             if get_origin(_type) != Annotated and len(get_args(_type)) != 2:
                 raise TypeError(f"Fields of a union need to be case or default, '{name}: {_type}' is not.")
+            if name in ['value', 'discriminator']:
+                raise TypeError(f"Field name '{name}' is disallowed in a union, perhaps '_{name}' was meant?")
 
             _type = get_args(_type)[1]
             if isinstance(_type, types.case):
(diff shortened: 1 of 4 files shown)


poetinger (approved):

FirebarSim:
Verified against OARIS 3.0 IDL that the Python version now warns about properties called value and suggests rename to _value

Issue #221, opened 2023-08-31
pip installation fails in Python 3.11


clunietp:
I'm attempting to install cyclonedds-python with the included binaries per the README.  

Reproducer Dockerfile:
    
    FROM python:3.11-bullseye
    RUN python3 -m pip install cyclonedds

Result:

    Collecting cyclonedds
    Downloading cyclonedds-0.10.2.tar.gz (156 kB)
       ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 156.8/156.8 kB 274.8 kB/s eta 0:00:00
    Installing build dependencies: started
    Installing build dependencies: finished with status 'done'
    Getting requirements to build wheel: started
    Getting requirements to build wheel: finished with status 'error'
    error: subprocess-exited-with-error
    
    × Getting requirements to build wheel did not run successfully.
    │ exit code: 1
    ╰─> [1 lines of output]
        Could not locate cyclonedds. Try to set CYCLONEDDS_HOME or CMAKE_PREFIX_PATH
        [end of output]

Installation seems to work fine if I use the `python:3.10-bullseye` image instead.

eburdette0:
I also ran into this issue on 3.11

LY1806620741:
Can anyone make a wheel package? I guess that only need to modify the python version 3.10->3.11.

eboasson:
@LY1806620741 

Finally, there's been some real effort on fixing the state of the python pipelines: https://dev.azure.com/eclipse-cyclonedds/cyclonedds-python/ thanks to the hard work of @trittsv. Now we can start looking at newer Python versions.

jc211:
Still having problems with this

ptfpinho23:
> Following instructions here solves the problem https://pypi.org/project/cyclonedds/0.9.0/#installing-with-pre-built-binaries but only with `bullseye` version

Can also confirm that this works on arch `6.10.11`

m-decoster:
Are there any plans to build official wheels for 3.11?

eboasson:
I'm totally fine with updating the set of wheels (it seems to do it for 3.11 on Ubuntu 22.04 but not for macOS or Windows), and I suppose there's not much point in continuing to build wheels for 3.7 or 3.8. I'm not much of a Python user myself, and I usually build it from the sources so I don't really notice.

What range of versions would you consider sensible these days?

m-decoster:
[...]

PR #248, opened 2024-05-29
support py3.11 and py3.12

LY1806620741:
try to support  py3.11 and py3.12

https://github.com/eclipse-cyclonedds/cyclonedds-python/issues/221#issuecomment-2135548768

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -53,15 +53,30 @@ jobs:
       'Ubuntu 22.04 LTS with Python 3.11':
         image: ubuntu-22.04
         python-version: "3.11"
+      'Ubuntu 22.04 LTS with Python 3.12':
+        image: ubuntu-22.04
+        python-version: "3.12"
       'macOS 11 with Python 3.10':
         image: macOS-11
         python-version: "3.10"
+      'macOS 11 with Python 3.11':
+        image: macOS-11
+        python-version: "3.11"
+      'macOS 11 with Python 3.12':
+        image: macOS-11
+        python-version: "3.12"
       'Windows Server with Python 3.7':
         image: windows-latest
         python-version: "3.7"
       'Windows Server with Python 3.10':
         image: windows-latest
         python-version: "3.10"
+      'Windows Server with Python 3.11':
+        image: windows-latest
+        python-version: "3.11"
+      'Windows Server with Python 3.12':
+        image: windows-latest
+        python-version: "3.12"
       'Fuzzer on Ubuntu 22.04 LTS with Python 3.7':
         image: ubuntu-22.04
         python-version: "3.7"
@@ -77,6 +92,14 @@ jobs:
       'Fuzzer on Ubuntu 22.04 LTS with Python 3.10':
         image: ubuntu-22.04
         python-version: "3.10"
+        run-fuzzing: true
+      'Fuzzer on Ubuntu 22.04 LTS with Python 3.11':
+        image: ubuntu-22.04
+        python-version: "3.11"
+        run-fuzzing: true
+      'Fuzzer on Ubuntu 22.04 LTS with Python 3.12':
+        image: ubuntu-22.04
+        python-version: "3.12"
         run-fuzzing: true
   steps:
   - download: current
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ test-command = "pytest -s --verbose {package}/tests"
 test-extras = ["dev"]
 
 before-all = [
-    "git clone --branch master --depth 1 https://github.com/eclipse-cyclonedds/cyclonedds.git main",
+    "git clone --branch releases/0.10.x --depth 1 https://github.com/eclipse-cyclonedds/cyclonedds.git",
```
eboasson, line 29: We mustn't change the core `cyclonedds` repo branch to `releases/0.10.x` on the `master` branch of the Python binding!

pyproject.toml
```diff
@@ -50,7 +50,7 @@ before-build = [
 
 # By default there is no repair command on windows set, use delvewheel
 repair-wheel-command = [
-    "pip install delvewheel==0.0.18",
+    "pip install delvewheel==1.6.0",
```
eboasson, line 53: `0.0.18` to `1.6.0` ... I think I should try to understand what `delvewheel` does before merging this 🤔

eboasson (commented):
Thanks @LY1806620741, you're right, the CI and release build setup for the Python binding really needs updating to current versions. It is so woefully behind because I'm no Python user myself and so I tend to postpone updating CI stuff for it ...

In any case, your proposal is not quite right yet, but I think it is quite close. Let's try to make this happen.

pyproject.toml
LY1806620741: `Delvewheel` is a command-line tool that modifies Python wheel packages and injects dependency dynamic libraries to make them independently bundled.

I built the Windows x64 version on my GitHub pipeline and it looks fine.
https://github.com/LY1806620741/unitree_sdk2_python/blob/37967b640de394992fc7bf06147daf856213a38b/.github/workflows/python-package.yml#L48

In the end, it injected a complete DLL file. see https://github.com/LY1806620741/unitree_sdk2_python/releases/tag/3.11

well, `delvewheel` 0.0.18 supports up to Python 3.9.see https://pypi.org/project/delvewheel/0.0.18/

at 1.0.0,it Label project as production/stable. https://github.com/adang1345/delvewheel/blob/master/CHANGELOG.md

pyproject.toml
LY1806620741: I think that we needed to add version 0.10.2  for cyclonedds-python, but there was a change in the main branch of C development. In fact, it is already a new version, so I am a bit confused about what to do

LY1806620741: I saw a bug in the main branch of Cyclonedds C version 0.10.3, which caused the build to fail, see https://dev.azure.com/eclipse-cyclonedds/cyclonedds-python/_build/results?buildId=6119&view=results

LY1806620741: It is possible to update and wait for a stable version of 0.10.3 before building a more secure one

azure-pipelines.yml
```diff
@@ -53,15 +53,30 @@ jobs:
       'Ubuntu 22.04 LTS with Python 3.11':
         image: ubuntu-22.04
         python-version: "3.11"
+      'Ubuntu 22.04 LTS with Python 3.12':
```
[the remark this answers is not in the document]
LY1806620741: Yesterday I installed Centos7, and found the latest rpm package is Python 3.6.This means that there may some people still using the old version

pyproject.toml
LY1806620741: @eboasson hi, I build python 3.11+ version for cyclonedds-python3.10.2、cyclonedds3.10.2, Can you confirm and upload it to the pypi library? The build script is open source, and the entire execution is in the github container. I tested that it is available on both windows and linux.

https://github.com/LY1806620741/unitree_sdk2_python/actions/runs/9350099214

jc211:
Any chance of this being merged soon? I'm currently keeping all my python environments on python 3.10 so that cyclonedds is only one pip install way. Would love to be able to upgrade :)

LY1806620741:
> Any chance of this being merged soon? I'm currently keeping all my python environments on python 3.10 so that cyclonedds is only one pip install way. Would love to be able to upgrade :)

I see https://pypi.org/project/cyclonedds/ support python3.10.
If you need python3.11+，you can use my builded https://github.com/LY1806620741/cyclonedds-python/releases/tag/0.10.2.

the way is download then `pip install filename`

PR #254, opened 2024-06-04
fix master nightly job

trittsv:
This PR fixes the nightly wheel build job.
And a other bug in the wheel build that on windows the wrong build type was used.

@eboasson ready for review :)


eboasson (approved):
Looks good (to my untrained eye), now lets see what happens when we let the "official" pipeline run 🙂

PR #261, opened 2024-07-02
Check well-formedness of CDR before extracting key

eboasson:
The python serializer doesn't detect that in:

    enum E { A, B, C }
    union T switch(E) { ... }
    union U switch(boolean) { case true: T value }
    struct S { U m }
    write (S(m=T(discriminant=B, value=...)))

there's an entire union missing in the value to be written, and because it can't find "B" in the list of labels in U (i.e., "B" is not in "[True]") it serializes "B" and skips "value".  It so happens that "B"'s serialization matches that of "True" (on a little endian machine, but there are variants where endianness doesn't matter), and so "dds_stream_extract_key" will take the "true" case and try to deserialize T.  It doesn't handle malformed input, so bad things happen.

I don't know how to prevent this type confusion in the Python serializer, but I do know how to check for malformed input.

### DIFF
--- a/clayer/pysertype.c
+++ b/clayer/pysertype.c
@@ -1,5 +1,4 @@
-/* -*- mode: c; c-basic-offset: 4 -*-
- *
+/*
  * Copyright(c) 2021 to 2022 ZettaScale Technology and others
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


clayer/pysertype.c
```diff
@@ -120,47 +119,60 @@ static ddspy_serdata_t *ddspy_serdata_new(const struct ddsi_sertype* type, enum
...
 
-static void ddspy_serdata_populate_key(ddspy_serdata_t* this)
-{
-    if (csertype(this)->keyless) {
-        this->key = NULL;
-        this->key_size = 0;
-        this->key_populated = true;
+static bool ddspy_serdata_populate_key(ddspy_serdata_t* this)
+{
+    const uint32_t xcdr_version = this->is_v2 ? DDSI_RTPS_CDR_ENC_VERSION_2 : DDSI_RTPS_CDR_ENC_VERSION_1;
```
dpotman, line 130: Looks like the keyless variant is not handled correctly, that results in a key with only the CDR header.

dpotman (commented):
Would be nice to eventually fix this in the serializer, but adding a check (normalize) before key extraction seems to be a good way to prevent most of the trouble. 

Most of the changes look fine to me, but I think the keyless variant is not handled correctly in `ddspy_serdata_populate_key`. 

A small nitpick: the white spacing is inconsistent in a few places in this file (it caught my attention because of the Emacs variable on the first line 😄)

dpotman (approved):
Looks good. Only remark is that for keyless types we now do the key extraction (which is a no-op in that case) and allocating a key buffer for just the header, but I think the performance impact of that is very limited, so approving this PR.

PR #273, opened 2024-10-31
Only serialize key when C function asks for a key

### LINKED ISSUE #269: Instance Registration and Disposal fails
When calling register_instance and passing in a sample of a type with defined keys the returned handle is a constant per run (i.e. run 1 always returns 123456789, run 2 always returns 987654321).

When calling dispose_instance_handle(instance_handle) a sample dispose is sent but for a different key than that of any of the registered instances.

When calling dispose_instance(sample) a sample dispose is sent but for a different key than that of the sample.

The samples being registered are OMG OARIS.

eboasson:
For some operations, the DDS API requires a key instead of a sample ("dispose" is one of those), and does this by invoking the "ddsi_serdata_from_sample" operation with kind set to "SDK_KEY".

The Python binding is a bit unusual in that it serializes the data before calling into C because that allows using Python code to serialize the data without having to call back into Python code from C or being forced to interpret Python objects from C.  A mismatch is possible, and this causes failures if the serialized form of the key is not a prefix of the serialized form of the full sample.  For example:

    @dataclass
    class S(idl.IdlStruct, typename="S"):
        v: int
        k: int
        key("k")

    ih0 = dw.register_instance(S(0, 1))

Will misinterpret the bytes and treat the key value as 0 instead of as 1.

It can be dealt with in two ways:

* One is to continue always serializing a full sample in Python, fixing it up in the serdata implementation if a serialized key is requested.  This is small change in the Python "clayer";
* The other is to serialize a key when a key is expected. This is somewhat fragile, but the list of operations is small and the API rather stable.

This commit implements the second option, and mostly because it means the serializer will not touch the non-key fields.

Intended to fix #269

### DIFF
--- a/cyclonedds/idl/_main.py
+++ b/cyclonedds/idl/_main.py
@@ -129,7 +129,7 @@ class IDL:
         # impossible for another thread to observe a partially populated self
         self._populated = True
 
-    def serialize(self, object, use_version_2: bool = None, buffer=None, endianness=None) -> bytes:
+    def serialize(self, object, use_version_2: bool = None, buffer=None, endianness : Endianness = None, serialize_kind: SerializeKind = SerializeKind.DataSample, prepend_header: bool = True) -> bytes:
         if not self._populated:
             self.populate()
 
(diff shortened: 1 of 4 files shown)


cyclonedds/sub.py
```diff
@@ -371,7 +371,7 @@ def wait_for_historical_data(self, timeout: int) -> bool:
         raise DDSException(ret, f"Occured while waiting for historical data in {repr(self)}")
 
     def lookup_instance(self, sample: _T) -> Optional[int]:
-        ret = ddspy_lookup_instance(self._ref, sample.serialize())
+        ret = ddspy_lookup_instance(self._ref, sample.serialize_key())
```
dpotman, line 374: I'd expect that this serialize also needs padding `\0`s, similar to `pub.lookup_instance`?

dpotman (commented):
Looks good! I have only two minor points noted below.

cyclonedds/sub.py
eboasson: Actually, no, and `pub.lookup_instance` doesn't really need it either. This padding business is a bit messy ...

It comes from the requirement that submessage headers in RTPS message are aligned to a multiple of 4 bytes and that data payloads start at an multiple of 4 bytes as well, which means that data (may) need to be padded out to a multiple of 4 bytes, too.

For better or for worse, I chose to include that padding in the serialized representation in the `serdata` in the C version, from where it metastasised to C++ and Python. The idea was that this allows a DATA submessage to be a pair of blobs in `sendmsg` (one containing just the message header, one the payload), instead of a triplet of blobs (a 3rd one for padding). Of course there are other options: in hindsight, it would've been better to *prepend* 4 bytes to the message header, because that padding at the end of the payload is highly annoying and trips everyone up — that's including me 😳😱🤦

(I think I can actually still make that change without causing any problems for a `serdata` implementation that does pad it out. It would just eliminate the need to do so.)

Anyway, the only place where it matters is in sending the data out on the wire. The "lookup instance" operation by definition doesn't do that, therefore it can't *require* the padding. It does make some sense to just pad everything out in the `pub` side of things, of course.

The keys do need to be the same because those get compared, but the Python binding (like the C code) stores a serialised key separately from the data, and it obtains it by calling `dds_stream_extract_key_from_key` or `dds_stream_extract_key_from_data`.

I think it might make the most sense to leave the padding bit here as it is now, and eliminate this weird padding instead.

cyclonedds/idl/_main.py
```diff
@@ -209,45 +209,14 @@ def _deserialize(self, data, has_header=True, use_version_2: bool = None, deseri
...
-
-        if self.version_support.SupportsBasic & self.version_support:
-            use_version_2 = False if use_version_2 is None else use_version_2
-        else:
-            # version 0 not supported
-            if use_version_2 is not None and not use_version_2:
-                raise Exception("Cannot encode this type with version 0, contains xcdrv2-type structures")
-            use_version_2 = True
-
-        if self.keyless:
```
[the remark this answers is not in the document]
eboasson: I concluded that it was really just an optimisation: if there are no keys, then `self.v{2,0}_machine.serialize` skips all the fields and yields an empty output. I figured the optimisation wasn't really worth the bother, because one doesn't really serialise keys all that often anyway.

I'll have another look, see if I still agree with my original analysis.

PR #281, opened 2024-12-09
Add set_topic_filter & matched_pub, sub & status getters & statistics

eboasson:
This is really just #43 repackaged to match the current status of the Python binding.

There are some conflicts, the solution of all of them could have been trivial but I instead decided to eliminate some code duplication in the construction of `DcpsEndpoint` objects. I also chose to use a merge commit, it is bad enough that @YixianLv 's PR was forgotten for 3 years, the least I can do is to build upon it rather than mess with it.


dpotman (approved):
Although I don't have in-depth knowledge of the Python binding, I performed a sanity check on the changes in this PR, and everything looks good to me!

PR #284, opened 2025-02-13
Fix invalid syntax in cli for python3.7

AlexisPotvinARA:
Python 3.7 does not support unpacking without parentheses in return statements


eboasson (approved):
LGTM, Thanks @AlexisPotvinARA !

PR #285, opened 2025-03-26
Inherit member ids from base for dyn disc type

eboasson:
This adds the the member ids from the base type to the member ids of a struct when dynamically constructing a struct type from a TypeObject.

Without this step, the inherited fields end up getting member id -1 somewhere in the of the deserializer and cause the inherited fields to not be found during deserialization.  Because of this, none of the inherited fields will be deserialized and a must-understand field (which includes key fields) among them causes a deserialization error.


dpotman (approved):
LGTM!

PR #293, opened 2025-05-18
feature/performance-improvement-readtake

trittsv:
The implementation of take/read is very slow because of `readtake_pre`, it allocates `MAX_SAMPLE_SIZE`-samples which takes forever.
Every `reader.read(MAX_SAMPLE_SIZE)` call takes around 700 ms.

It made it impossible to implement the "shape-tail" in CycloneDDS-Insight because it was too slow to get all samples with a fluent shape movement.

This PR replaces the `dds_read/take` methods with `dds_read/take_with_collector` and allocates only the amount of memory which is actually needed.

@eboasson could you have a look?

### DIFF
--- a/clayer/pysertype.c
+++ b/clayer/pysertype.c
@@ -67,6 +67,14 @@ typedef struct ddspy_sample_container {
   size_t usample_size;
 } ddspy_sample_container_t;
 
+typedef struct {
+  ddspy_sample_container_t *containers;
+  dds_sample_info_t *sample_infos;
+  size_t max_samples;
+  size_t count;
+  size_t capacity;
+} collector_state_t;
+
 static inline ddspy_sertype_t *sertype (ddspy_serdata_t *this)
 {
   return (ddspy_sertype_t *) (this->c_data.type);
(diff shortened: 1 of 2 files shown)


clayer/pysertype.c
```diff
@@ -67,6 +67,14 @@ typedef struct ddspy_sample_container {
   size_t usample_size;
 } ddspy_sample_container_t;
 
+typedef struct {
+  ddspy_sample_container_t *containers;
+  dds_sample_info_t *sample_infos;
+  size_t max_samples;
```
eboasson, line 73: Do you really need this `max_samples`? `dds_take_with_collector` (&c.) already has a limit, and so I would not expect `collector_callback_fn` to ever get called with `count >= max_samples` on entry.

clayer/pysertype.c
```diff
@@ -1063,76 +1071,124 @@ static inline uint32_t check_number_of_samples (long long n)
...
 {
-  if (sts < 0)
-    return PyLong_FromLong ((long)sts);
+  collector_state_t *state = (collector_state_t *)arg;
 
-  PyObject *list = PyList_New (sts);
-  for (int32_t i = 0; i < sts; ++i)
+  if (state->count >= state->max_samples)
+    return DDS_RETCODE_OK;
+
```
eboasson, line 1108: So this I think isn't needed ...
```suggestion
```

clayer/pysertype.c
```diff
@@ -1063,76 +1071,124 @@ static inline uint32_t check_number_of_samples (long long n)
...
-    dds_free (container[i].usample);
+    // Grow allocation, this ensures amortized linear growth while keeping allocation calls minimal.
+    // Doubling gives exponential growth this makes adding N items only require log2(N) reallocations - efficient!
+    size_t new_capacity = state->capacity ? state->capacity * 2 : 8;
+    
+    void *new_containers = dds_realloc(state->containers, new_capacity * sizeof(ddspy_sample_container_t));
+    void *new_infos = dds_realloc(state->sample_infos, new_capacity * sizeof(dds_sample_info_t));
+
+    if (!new_containers || !new_infos)
+      return DDS_RETCODE_OUT_OF_RESOURCES;
```
eboasson, line 1119: This I think is fine: it simply means the take/read operation stops early, and then with the modified condition in `readtake_post` it will return `DDS_RETCODE_OUT_OF_RESOURCES` if this is on the initial sample, or the data it did collect if it is on a later one

clayer/pysertype.c
```diff
@@ -1063,76 +1071,124 @@ static inline uint32_t check_number_of_samples (long long n)
...
+    state->sample_infos = new_infos;
+    state->capacity = new_capacity;
   }
-  dds_free (info);
-  dds_free (container);
-  dds_free (rcontainer);
-  return list;
+
+  if (!serdata_to_sample (serdata,  &state->containers[state->count], NULL, NULL))
+    return DDS_RETCODE_ERROR;
```
eboasson, line 1127: Given the implementation `serdata_to_sample` on line 352, this really just means out-of-memory. So I think it makes the most sense to here also do:
```suggestion
    return DDS_RETCODE_OUT_OF_RESOURCES;
```

clayer/pysertype.c
```diff
@@ -1063,76 +1071,124 @@ static inline uint32_t check_number_of_samples (long long n)
...
+
+  dds_return_t sts = readtake(
+    reader,
+    (uint32_t)N,
+    DDS_HANDLE_NIL,
+    mask,
+    collector_callback_fn,
+    &state);
+
+    return readtake_post((int32_t)sts, &state);
```
eboasson, line 1162:
```suggestion
  return readtake_post((int32_t)sts, &state);
```

clayer/pysertype.c
```diff
@@ -1063,76 +1071,124 @@ static inline uint32_t check_number_of_samples (long long n)
...
-static bool readtake_pre (long long N, uint32_t *Nu32, dds_sample_info_t **info, ddspy_sample_container_t **container, ddspy_sample_container_t ***rcontainer)
+static PyObject *readtake_post (int32_t sts, collector_state_t *state)
 {
-  if (!(*Nu32 = check_number_of_samples (N)))
-    return false;
-
-  *info = dds_alloc (sizeof (**info) * *Nu32);
-  *container = dds_alloc (sizeof (**container) * *Nu32);
-  *rcontainer = dds_alloc (sizeof (**rcontainer) * *Nu32);
+  if (sts < 0)
```
eboasson, line 1076: If some samples have been collected before the error occurred, this will leak. An error during the collection step is a bit of a problem: some states changes occur in the process and it can't roll it back. The options I see are:

* free the memory, return an error and accept that some samples are lost forever (or marked as read but never actually seen by the application)
* return the collected samples and pretend nothing went wrong, on the (reasonable) assumption that the application will try again for the remainder
* return the collected samples while (somehow) recording an error occurred, and returning an error the next time round

I wouldn't go for the third one, it is terribly confusing to get a seemingly correct result from a failing operation, and then an error code much later when there is no apparent cause.

The errors that can occur in practice are running out of memory or failing to deserialize the data. The first is exceedingly unlikely, the second can also be done by just dropping only the failing sample.

So here I would do:
```suggestion
  if (sts < 0 && state->count == 0)
```

eboasson (commented):
Thank you so much @trittsv ! This is something that I wanted to do from the day I added the `..._with_collector` functions but never actually got around to doing.

I do think there is a memory leak on an out-of-memory condition now. For a little while I was tempted to merge it as-is because I am so happy about it, but I think it is better to try to fix that first. (And I think it is trivial to do so.)

clayer/pysertype.c
trittsv: That is true, that was may a bit over protective, i think we can trust `dds_take_with_collector` in that point.
I removed it.

clayer/pysertype.c
trittsv: see comment above, i removed it.

clayer/pysertype.c
trittsv: indention fixed

clayer/pysertype.c
trittsv: good catch, i agree with the suggestion to return what we have already successfully collected, the user will happily collect the missing in the next round.

trittsv:
@eboasson thank you very much for your review. I implemented the points you commented. Could you have a look again?

eboasson (approved):
LGTM!

PR #329, opened 2026-04-29
`cli:` show DCPSTopic data if available

Splinter1984:
If topic discovery is enabled fill the discovery endpoints with discovered topic entities (applies for `cyclonedds` and `ddsls` cli-tools).


eboasson (approved):
Thanks @Splinter1984 ! I tried it with a case where the topic QoS differed from the endpoint QoS and it works like a charm :)

PR #335, opened 2026-07-29
Update assertions with removed variables

nmarticorena:
Hi all, when building `cyclonedds-py` natively in `osx` I found the following issue:

```
clayer/pysertype.c:493:11: error: use of undeclared identifier 'le_key'
  assert (le_key != NULL);
          ^

clayer/pysertype.c:494:11: error: use of undeclared identifier 'le_keysz'
  assert (le_keysz > 0);
          ^
```

The failure may be hidden in builds that define NDEBUG, because the preprocessor removes the two assertions before compilation.

The stale identifiers were left behind when the internal key representation was changed in commit 6428be8c53f627f43e4529918a13dc25e7c6fb07. The remaining code already uses the replacement fields:

cserdata(d)->key
cserdata(d)->key_size

This patch was tested in the `conda-forge` build https://github.com/conda-forge/cyclonedds-py-feedstock/pull/1


eboasson (commented):
Thank you @nmarticorena. Happy to merge it like it is, but before I can do so, you need to sign the [ECA](https://www.eclipse.org/legal/eca/). Please let me know!

nmarticorena:
Thanks for the quick review @eboasson , it should be done I did it just after I open this pr, please let me know if it works :)

eboasson (approved):
Indeed! Thanks @nmarticorena 😀

PR #341, opened 2026-09-14
Fix keyless topic dispose

### LINKED ISSUE #340: Disposing a keyless mutable XCDR2 type raises struct.error
Disposing a keyless mutable XCDR2 type causes DataReader.take() in Python to raise:
```
struct.error: unpack_from requires a buffer of at least 8 bytes for unpacking 4 bytes at offset 4 (actual buffer size is 4)
```
Minimal reproducing IDL:
```
module DisposeRepro {
    @mutable
    @data_representation(XCDR2)
    @topic
    struct Message {
        string value;
    };
};
```
Steps to reproduce (for example in CycloneDDS-Insight):

1. Create a reader and writer for this type.
2. Write a sample. The reader receives it successfully.
3. Dispose the instance, or delete the writer.
4. The Python reader raises the exception above in take().

Adding `@key` to `string value;` makes disposal work. I also reproduced the same error in the Python reader when disposing from writers implemented in other languages.

The failure occurs in `deserialize_key()`: the mutable XCDR2 decoder attempts to read a four-byte struct-size field at offset 4, but the supplied buffer is only four bytes long.

Expected behavior: the reader returns the lifecycle notification as an InvalidSample without raising an exception.

@eboasson, could you take a look?

eboasson:
- Serialize keyless topic keys without a payload.
- Preserve the complete operation stream when initializing dynamically generated CDR descriptors.
- Ignore locally generated Python wheel archives.

## Keyless topic serialization

Keyless final, appendable, and mutable types now produce an empty key payload for both XCDR1 and XCDR2.

The public `serialize_key()` API still includes the four-byte representation header, while the headerless serialization used by the C layer is empty. Key deserialization consumes the header when present and returns a default-initialized instance without attempting to deserialize a nonexistent key.

This also fixes handling of invalid samples, such as dispose notifications, for keyless mutable topics.

## XCDR1 optional members

The Python binding initialized dynamically reconstructed CDR descriptors using the legacy `dds_cdrstream_desc_init()` function. That function derives the operation-stream length and can omit metadata appended after the normal and key operations.

For types with optional members, this dropped the `DDS_OP_MID` table required to interpret XCDR1 parameter headers. Consequently, otherwise valid CDR was rejected during `DataWriter.write()` with `DDS_RETCODE_BAD_PARAMETER`.

Use `dds_cdrstream_desc_init_with_nops()` and pass `desc.m_nops` so the complete operation stream is retained. Initialization errors are now propagated as well.

A deterministic regression test covers both present and absent XCDR1 optional values.

## Housekeeping

Add `*.whl` to `.gitignore` so locally built or downloaded Python wheels are not reported as untracked files.

Fixes #340


trittsv:
@eboasson thanks a lot :)

